PR #2, opened 2020-03-16
Version 0.9.0

curquiza:


src/Index.php
```diff
@@ -53,14 +49,14 @@ public function getDocuments($options = null)
         return $this->httpGet('/indexes/'.$this->uid.'/documents', $options);
     }
 
-    public function addOrReplaceDocuments($documents)
+    public function addDocuments($documents, $primary_key = null)
```
qdequele: We're done with addOrReplace?

src/Index.php
```diff
@@ -53,14 +49,14 @@ public function getDocuments($options = null)
...
     {
-        return $this->httpPost('/indexes/'.$this->uid.'/documents', $documents);
+        return $this->httpPost('/indexes/'.$this->uid.'/documents', $documents, ['primaryKey' => $primary_key]);
     }
 
-    public function addOrUpdateDocuments($documents)
+    public function updateDocuments($documents, $primary_key = null)
```
qdequele: We're done with addOrUpdate?

tests/SearchTest.php
```diff
@@ -25,10 +28,16 @@ public static function setUpBeforeClass(): void
             ['id' => 4,    'title' => 'Harry Potter and the Half-Blood Prince', 'comment' => 'The best book'],
             ['id' => 42,   'title' => 'The Hitchhiker\'s Guide to the Galaxy'],
         ];
-        static::$index->addOrUpdateDocuments($documents);
+        static::$index->updateDocuments($documents);
         usleep(10 * 1000);
```
qdequele: Is it difficult to wait for update with a loop on getOneUpdate(updateId)?

src/Index.php
curquiza: Yes, in all the SDK we use `add_documents` and `update_documents` instead of the longer ones `add_or_replace` and `add_or_update` to be clearer. One method for one usage.

tests/SearchTest.php
curquiza: Yes, it's an alternative to the `usleep` to do a sync method of each method for the test with this kind of loop. But now it would add me a looooooot of work to do it in all the SDKs, because the PHP SDK is not the only one concerned.

src/Index.php
curquiza: Yes, in all the SDK we use `add_documents` and `update_documents` instead of the longer ones `add_or_replace` and `add_or_update` to be clearer. One method for one usage.

tests/SearchTest.php
qdequele: You can also create only one method who take a update_id and do a loop. like that: 
```php
class Index extends HTTPRequest
{
    public function waitUpdateId($update_id)
    {
        while (1) {
            $res = $this->getUpdateStatus($update_id);
            if ($res['status'] == 'processed' || $res['status'] == 'failed') {
                return $res;
            }
            usleep(500);
        }
    }
}
```

I can propose a PR if you want!

curquiza: Ok as you want :)
Can you do it after we merged this one? It would be clearer to review :)

qdequele: yes!

PR #19, opened 2020-04-21
Refactored test suite to make sure the tests are running isolated

ppshobi:
Hi, This is obviously a big pull request. Please take time to review it especially because this is the test suite. But I have made changes in small chunks so that the review process will be easier. That is also the reason you can see a big commit list.

I will give you an idea about the changes I have done in this PR.

- There is a base TestCase.php class which all the test classes should extend and its a place we can have the basic bootstrapping of our test suite
- Then I have gone through each test classes and removed/replaced the static methods by making use of `constructor()` and `setup()` methods as necessary and isolated each test methods to be run independently of each other. Then made sure the entire test class passes and then the whole suite. Then move on to the next class.
- One or two extra tests were added and removed as necessary. Some method extraction refactoring was done on a few tests as it would cut the duplication.

### DIFF
--- a/scripts/tests.sh
+++ b/scripts/tests.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 echo 'Setting FDs limit to 1000'
-ulimit -n 1000
+ulimit -Sn 1000
 
 echo "Launching tests..."
 vendor/bin/phpunit --color tests/
--- a/src/Client.php
+++ b/src/Client.php
@@ -21,6 +21,13 @@ class Client extends HTTPRequest
         return $this->indexInstance($uid)->delete();
     }
 
+    public function deleteAllIndexes()
+    {
+        foreach ($this->getAllIndexes() as $index) {
+            $this->deleteIndex($index['uid']);
+        }
+    }
+
     public function getIndex($uid)
     {
         return $this->indexInstance($uid);
--- a/composer.json
+++ b/composer.json
@@ -23,10 +23,7 @@
     "autoload-dev": {
         "psr-4": {
             "Tests\\": "tests/"
-        },
-        "files": [
-            "tests/utils.php"
-        ]
+        }
     },
     "require-dev": {
         "phpunit/phpunit": "^8.5",
--- a/src/Index.php
+++ b/src/Index.php
@@ -106,13 +106,14 @@ class Index extends HTTPRequest
 
     public function search($query, $options = null)
     {
-        if (isset($options)) {
-            $paramters = array_merge(['q' => $query], $options);
+        if ($options) {
+            $options = $this->flattenOptions($options);
+            $parameters = array_merge(['q' => $query], $options);
         } else {
-            $paramters = ['q' => $query];
+            $parameters = ['q' => $query];
         }
 
-        return $this->httpGet('/indexes/'.$this->uid.'/search', $paramters);
+        return $this->httpGet('/indexes/'.$this->uid.'/search', $parameters);
     }
 
     // Stats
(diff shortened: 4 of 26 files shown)


curquiza:
Related to #23

ppshobi:
I think the PR is ready to review. Please let me know what you think. I have some other refactorings in mind, but it would require changes in the `src/` so I have refrained myself from doing it at least in this PR.

src/Client.php
```diff
@@ -21,6 +21,13 @@ public function deleteIndex($uid)
         return $this->indexInstance($uid)->delete();
     }
 
+    public function deleteAllIndexes()
```
curquiza, line 24: Since it's a method we provide now for our users, we should test it 🙂
- test when there is no index
- test when there are indexes

ppshobi: closed with 67adb2f

scripts/tests.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+echo 'Setting FDs limit to 1000'
+ulimit -n 1000
+
+echo "Launching tests..."
+vendor/bin/phpunit --color tests/
```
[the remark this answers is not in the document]
Kerollmops: `ulimits` are setup for only the pid and the childs of this pid IIRC.
@curquiza The question I have is why don't you use the `-S` (soft?) flag?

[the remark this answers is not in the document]
curquiza: @Kerollmops my bad!!

scripts/tests.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+echo 'Setting FDs limit to 1000'
+ulimit -n 1000
```
curquiza, line 4:
```suggestion
ulimit -Sn 1000
```

curquiza: Sorry, my mistake

curquiza:
Related to [this issue](https://github.com/meilisearch/integrations-guides/issues/11).

bidoubiwa:
Is it possible to add two tests in search ?
Those tests are made to see if the SDK handles correctly the fact that some search parameters can either be an array of strings or a single string.

## Test 1
```javascript
{
        limit: 5,
        offset: 0,
        attributesToRetrieve: ['id', 'title'],
        attributesToCrop: ['id', 'title'],
        cropLength: 6,
        attributesToHighlight: ['id', 'title'],
        filters: 'title = "Le Petit Prince"',
        matches: true,
}
```

You should have these keys
```php
$this->assertArrayHasKey('_matchesInfo', $response["hits"][0]);
$this->assertArrayHasKey('title', $response["hits"][0]['_matchesInfo']);
$this->assertArrayHasKey('_formatted', $response["hits"][0]);
```

But not these keys:

```php
$this->assertArrayNotHasKey('comment', $response["hits"][0]);
$this->assertArrayNotHasKey('comment', $response["hits"][0]['_matchesInfo']);
```

Try those value
```php
$this->assertSame('Petit <em>Prince</em>', $response["hits"][0]["_formatted"]["title"]);
```

## test 2
```javascript
{
        limit: 5,
        offset: 0,
        attributesToRetrieve: "*",
        attributesToCrop: "*",
        cropLength: 6,
        attributesToHighlight: "*",
        filters: 'title = "Le Petit Prince"',
        matches: true,
}
```

You should have these keys
```php
$this->assertArrayHasKey('_matchesInfo', $response["hits"][0]);
$this->assertArrayHasKey('_formatted', $response["hits"][0]);
$this->assertArrayHasKey('comment', $response["hits"][0])
$this->assertArrayHasKey('comment', $response["hits"][0]["_formatted"])
$this->assertArrayHasKey('title', $response["hits"][0]['_formatted']);
$this->assertArrayHasKey('title', $response["hits"][0]['_matchesInfo']);
$this->assertArrayHasKey('comment', $response["hits"][0]['_matchesInfo'])
```

Try those value
```php
$this->assertSame('Petit <em>Prince</em>', $response["hits"][0]["_formatted"]["title"]);
```

Tests could be improved but this assure us that the strings are handled correctly!

curquiza:
Need rebasing, and we are finally done @ppshobi 🙂 🎉

ppshobi:
Done 👍

curquiza:
Perfeeeeeeect 👌 Thanks again @ppshobi!

PR #46, opened 2020-06-29
Add getting started code samples

curquiza:
Closes partially #34

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -1,3 +1,8 @@
+# This code-samples file is used by the MeiliSearch documentation
+# Every example written here will be automatically fetched by
+# the documentation on build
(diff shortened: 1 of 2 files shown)


.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,111 @@
...
+documents_guide_add_movie_1: |-
+search_guide_1: |-
+search_guide_2: |-
+getting_started_create_index_md: |-
+  ```bash
+  $ composer require meilisearch/meilisearch-php
+  ```
+
+  ```php
+  <?php
```
bidoubiwa, line 81: This opening tag is only present in this sample and is not closed with `?>`.

curquiza: No need to close with `?>`

PR #47, opened 2020-06-29
Implement getOrCreateIndex method

emulienfou:
Implements the `getOrCreateIndex` method, as specified in [this discussion](https://github.com/meilisearch/integration-guides/issues/2)

* Add `getOrCreateIndex` method
* Add tests
* Update Readme with examples

> Related to issue #26

PR #49, opened 2020-06-30
Fix/request exception

emulienfou:
This a new fix related to class `MeiliSearch\Exceptions\HTTPRequestException`

I had a case where I configured a bad client URL and the code returned a string in variable `$http_body` instead of array and the code returned this error message: `array_key_exists() expects parameter 2 to be array, string given`.

I also in the same PR fix other tests classes who are not compatible with PHP7.2.

I added a new test for the **HTTPRequestException** class to check my changes using a bad client url.

> This PR code depends of the previous one: https://github.com/meilisearch/meilisearch-php/pull/47

### DIFF
--- a/src/Exceptions/HTTPRequestException.php
+++ b/src/Exceptions/HTTPRequestException.php
@@ -2,6 +2,9 @@
 
 namespace MeiliSearch\Exceptions;
 
+use function array_key_exists;
+use function is_array;
+
 class HTTPRequestException extends \Exception
 {
     public $http_status = 0;
@@ -11,8 +14,9 @@ class HTTPRequestException extends \Exception
     public function __construct($http_status, $http_body, $previous = null)
     {
         $this->http_body = $http_body;
-        if (isset($this->http_body) && array_key_exists('message', $this->http_body)) {
-            $this->http_message = $this->http_body['message'];
+        if (!empty($this->http_body)) {
+            $this->http_message = is_array($this->http_body) && array_key_exists('message', $this->http_body) ?
+                $this->http_body['message'] : $this->http_body;
         }
         $this->http_status = $http_status;
         parent::__construct($this->http_message, $this->http_status, $previous);
--- a/tests/Exception/HTTPRequestExceptionTest.php
+++ b/tests/Exception/HTTPRequestExceptionTest.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Tests\Exception;
+
+use MeiliSearch\Exceptions\HTTPRequestException;
+use Tests\TestCase;
+
+/**
+ * Class HTTPRequestExceptionTest.
+ */
+class HTTPRequestExceptionTest extends TestCase
+{
+    public function testBadClientUrl()
+    {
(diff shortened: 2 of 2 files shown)


devrck:
This PR can go also without depending on the previous i guess right?

emulienfou:
@devrck I create the branch from the #47 branch so I think we will need to wait to my other PR to be merged

src/Exceptions/HTTPRequestException.php
```diff
@@ -2,6 +2,9 @@
 
 namespace MeiliSearch\Exceptions;
 
+use function array_key_exists;
```
devrck, line 5: Maybe we can add this line into `.php_cs.dist` to get a more stable/concise way of doing it:

```php

$rules = [
    // ...
    'native_function_invocation' => ['include' => ['@compiler_optimized'], 'scope' => 'namespaced'],
]
```
What do you think?

curquiza: What this configuration does imply? 🙂

curquiza:
This PR needs to be rebased 🙂

src/Exceptions/HTTPRequestException.php
devrck: instead of doing the `use function array_key_exists;` and everybody can decide if they should or not do this one we can add it in the configuration and actually it will do `\array_key_exists()`.

The reason for even doing this is because of some performance gains.

An example https://github.com/scssphp/scssphp/issues/96 measurement done by a symfony maintainer.

curquiza: So interesting! Thanks @devrck for the explanation!

> Of course at big scale you feel it but in the day to day you won't even notice it.

Completely agree. If we can easily do improvement now, we take it!

devrck: Of course we can open a new PR and apply it in the whole codebase instead of putting more in this PR.

curquiza: @devrck
Are there other native functions we need to change in this package (by putting the `\` as prefix name)?
In this case, we need to do a new PR, but I just wanted to know 🙂

devrck: Here is one example https://github.com/meilisearch/meilisearch-php/blob/master/src/Index.php#L282 that is outside of @emulienfou PR and I think it should be done separate.

This is the full list https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.16/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L357

curquiza: @devrck
Are you ok to do a complete PR to add this improvement? 🙂
I'm going to merge this PR without.
If you can I will do it anyway when I have time 😉

devrck: Sure thing.

curquiza: I've just created an issue #53

PR #52, opened 2020-07-02
Added dependabot v1

devrck:
Keeping dependencies up-to-date.

https://dependabot.com/docs/config-file/

https://dependabot.com/

This of course can be installed on all clients and even on the core repository.

You also need to allow dependabot on the repo-level but I can definitely say it's really good for what it does.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,9 @@
+version: 1
+update_configs:
+    - package_manager: "php:composer"
+      directory: "/"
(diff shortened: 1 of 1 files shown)


curquiza:
Hello @devrck! Thanks for your PR.

But we already have dependabot on this package 🙂

Capture d’écran 2020-07-02 à 10 30 50

devrck:
Nice stuff :+1:

PR #54, opened 2020-07-02
Implementing native fn invocation

devrck:
* Applied to all code base 
* Optional: added declare_strict_types since we require PHP 7.2 :smile: (I can also remove it if we have a problem)
* Removed redundant directives from `.editorconfig` since they are provided from root `[*]`
* Moved linting before running tests.

Fixing https://github.com/meilisearch/meilisearch-php/issues/53

Issue #53, opened 2020-07-02
Use native_function_invocation to improve performance


curquiza:
See this [conversation](https://github.com/meilisearch/meilisearch-php/pull/49#discussion_r448798024) for complete details 🚀

ppshobi:
https://github.com/meilisearch/meilisearch-php/pull/56#issuecomment-653875138

PR #56, opened 2020-07-03
Apply PHPCS

devrck:
Tacke issue https://github.com/meilisearch/meilisearch-php/issues/53

Followup on https://github.com/meilisearch/meilisearch-php/pull/54


devrck:
:smile: I like when enforcing PHP7 and tests start failing :smile: some hidden bugs

devrck:
@curquiza, @ppshobi would want to re-write this to https://github.com/meilisearch/meilisearch-php/blob/master/src/Endpoints/Delegates/HandlesDocuments.php#L7

```php
<?php

namespace MeiliSearch\Endpoints\Delegates;

trait HandlesDocuments
{
    public function getDocument($documentId) // in my opinion this can be int, string or object with a __toString method 
    {
        $this->assertValid($documentId);
        //...   
    }

    
}
```

curquiza:
@devrck You're right it can be an int or a string. What do you think @ppshobi? And even @Guikingone if you are around? 😇

Guikingone:
Hi 👋 

Maybe it could be an option (on a dedicated PR?) to wrap the return of a document query within a `Document` object? 

This way, if the query fails, we can thrown an exception and if everything goes right, return an object? 

Maybe just something like this: 

```php
final class Document
{
    private $data;

    // ... Constructor

    public function getData(): array
    {
        return $this->data;
    }
}
```

The document object can have several shortcuts if needed 🤔

tests/Endpoints/DocumentsTest.php
```diff
@@ -59,7 +61,7 @@ public function testReplaceDocuments()
         $this->assertSame($replacement['title'], $response['title']);
         $this->assertFalse(array_search('comment', $response));
         $response = $index->getDocuments();
-        $this->assertCount(count(self::DOCUMENTS), $response);
+        $this->assertCount(\count(self::DOCUMENTS), $response);
```
ppshobi, line 64: Why there is a slash prefix  in `\count` calls??

src/Endpoints/Delegates/HandlesDocuments.php
```diff
@@ -4,11 +4,19 @@
...
+ */
 trait HandlesDocuments
 {
-    public function getDocument(string $document_id)
+    public function getDocument($documentId)
     {
-        return $this->http->get(self::PATH.'/'.$this->uid.'/documents/'.$document_id);
+        $this->assertValidDocumentId($documentId);
+
+        return $this->http->get(\sprintf('%s/%s/documents/%s', self::PATH, $this->uid, $documentId));
```
ppshobi, line 19: Here also the prefixed slash can be avoided. It looks dirty to me.

ppshobi:
> Hi 👋
> 
> Maybe it could be an option (on a dedicated PR?) to wrap the return of a document query within a `Document` object?
> 
> This way, if the query fails, we can thrown an exception and if everything goes right, return an object?
> 
> Maybe just something like this:
> 
> ```
> final class Document
> {
>     private $data;
> 
>     // ... Constructor
> 
>     public function getData(): array
>     {
>         return $this->data;
>     }
> }
> ```
> 
> The document object can have several shortcuts if needed 🤔

I have mixed feeling about this since, it will help us to have better types, but this document object will mostly act as a DTO object. And
1. All of the functionalities given by the DTO is available from an array except maybe syntactic sugar. 
2. During every request we will have to hydrate these DTOs from the meilisearch response. 
3. Depends on the response, this can be a heavy operation. (Typically search responses are small I agree, but this can be a case)
4. Also I believe we should give the end user the capability to change whatever they want, for which an array sounds better. 
5. Given a DTO we will have to provide methods to do trivial tasks such as `toArray`, `Iterate`, `get`,`set` etc.. which is an overhead for the library even though they are trivial.
6. Given all these can be done, the user might extend the class and add more methods to it since the functionalities doesn't suite his needs. But for such a user an array response will do more or less the same. He can construct DTO, even wrap the response around a fluent library such as `collection` from laravel.
7. Simple is better than complex

Guikingone:
Agree, in the end, the DX is more important than everything else. 

BTW, typing the `getDocument()` method should not be a problem IMO, if the API returns something else than an array, it could be a good idea to throw exceptions or log, this way, we can type the return a `: ?array` ? 🤔

ppshobi:
@devrck Its my bad that I haven't read the other conversation [here](https://github.com/meilisearch/meilisearch-php/pull/49#discussion_r448798024). As far as I understand this is supposed to bring in performance gains. But do we have any benchmarks that it has advantages in our case?

for example symfony rejected using FQN https://github.com/symfony/symfony/issues/21020 

and according to a post [here](https://decisions.typo3.org/t/should-we-use-fully-qualified-function-fqn-calls-like-is-array-in-the-core/392)  at typo3

- Pro
little performance gain

- Con
- makes code harder to read 
- Symfony decided against introducing it 
- existing benchmarks are very low level (testing function calls) no TYPO3 benchmarks were done, so the gain for real use case is unknown
- there is a workaround available - a composer plugin 14 which automatically changes code in vendor, and other tools which converts the code automaticaly
- it might break some testing code (where namespaces were used for mocking global functions)

unless we have clear performance gains, I think we should do this later. I would like to quote Mr. Donald Knuth.
> Premature optimization is the root of all evil

devrck:
Hi @ppshobi,

Thank you for taking the time to review :smile: .

The real quote is far longer and to be honest people misuse it and actually the idea is that if we spend a lot time doing micro-optimizations then yes it's not good but we have a tool for this, and we spend 0 effort on it:

> The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming

IMHO a lib should promote simple & performant code.

That being said I'm a fan of doing even the little stuff if I don't have to put effort of thinking about it. If symfony decided against it's not necessarily the way to move forward ( btw nicolas is the guy that promoted this :smile: but the community said no ), and regarding mocking a core function of php, that to me is fishy business.

If you guys say it's not worth doing it I'm also fine.

ppshobi:
Your contribution is definitely appreciated. And I didn't knew it was a thing to optimize native function calls to yield performance boost. Thank you for introducing this to me. Its great if we can have performance boost, but do we have enough pros to outweigh the cons? Are we facing performance issues? Also we could gain performance if we use single quotes instead of double quotes, but is that necessary?

Take another example about PHP JIT in php 8, it was hyped so much that the performance would be amazing, but real world benchmarking yielded more or less the same performance. At least as of now we do not have a benchmark to test our library against and we do not know how much performance it gains.

Mocking a native php function call is sketchy for me as well. Free performance sounds great. But It would have been cool if a release step automatically does this thing while packaging the library, but it seems like we have to do this at the codebase level. To me its a little sketchy even though a tool is doing this for us. Simple is better than complex and I feel like the prefixes makes the code harder to read, it's just me though. 

Community is you, me and others; lets drive this together. Thanks for your work man. @curquiza What do you think?

Guikingone:
@ppshobi 

> Take another example about PHP JIT in php 8, it was hyped so much that the performance would be amazing, but real world benchmarking yielded more or less the same performance. At least as of now we do not have a benchmark to test our library against and we do not know how much performance it gains.

Actually, JIT is really useful for librairies and internal PHP call, not for our applications as we mostly depends on the I/O, a good example of "JIT optimised" code is an image transformation library which don't rely on external calls. 

For the native functions, keep in mind that NOT every function need to be prefixed (mainly due to the fact that not every methods are handled internally by opcache), the list of methods handled and optimised by opcache is available here: https://github.com/php/php-src/blob/master/Zend/zend_compile.c#L3947. 

IMO, prefixing every method is a mistake but prefixing the right ones is a true optimisation path (IMO, we can easily use Blackfire on this SDK to identify bottlenecks which are bigger than native functions) and can lead to better code, better performances and "future proof" maintainability.

curquiza:
Hello guys! @ppshobi @Guikingone @devrck Sorry for the delay!

First of all, thanks for your involvement and your time! The PHP community is the most involved community in the MeiliSearch SDKs so far 😁 💪 

I'm not a PHP expert as you are guys, but I'm in charge of the organization of all the SDKs, and for me, there are 2 priorities:
1. The user experience of our package: the way of using it matters. We want our library to be easy to use! I like calling it "README-driven-development". I mean, give me the usage/getting-started of the feature, I would say if the Meili team is ok with that or not. Not need to be a PHP expert for this part 😉
2. The readability of our codebase so that everyone (expert or not) can contribute. Maybe the harder point to keep.

Because of these both points, I'm not really a big fan of [this suggestion](https://github.com/meilisearch/meilisearch-php/pull/56#issuecomment-653789902). But I understand that is not the main debate here 😉

About applying `native_function_invocation`, I didn't realize it could involve/need this kind of discussion. Sorry not to have asked for your opinion in the related issue before.
From my POV, including `'native_function_invocation' => ['include' => ['@compiler_optimized'], 'scope' => 'namespaced']` does not involve a lot of changes in our code, only the prefix `\` dash for some of the functions (not all of them), so it does not break the second point. But I might be wrong: maybe I don't see the other changes it would imply in the future. Do I forget something?

Anyway, as @ppshobi perfectly said, the Community is you guys, so maybe you could vote about `native_function_invocation`? (👍 or 👎 in my comment). I suggest the vote because I don't see any consensus for the moment 🙂 And I would rather not impose my decision since you seem better informed than me on the topic.

<hr>

@devrck I will do the review after the debate is closed 🙂 Sorry again for the delay! And most of all, thanks a lot for your work! 🚀
In any case, could you [rebase your branch](https://gist.github.com/curquiza/5f7ce615f85331f083cd467fc4e19398), please? 🙂

<hr>

That is no really related, but for those who don't know, here is the newborn in the MeiliSearch tools for the PHP community: the [Symfony bundle](https://github.com/meilisearch/meilisearch-symfony/) implemented by @emulienfou!!! 🎉 It's not pushed on Packagist yet, I'm on it this week.
Btw @emulienfou if you want to give your opinion on this debate, you are more than welcome 😇

curquiza:
It seems like everyone voted 🙂
Let's start applying this change then!

Keep in mind that **nothing is definitive in this package**: we are not stable yet, and if we realize this change will involve more complexity than we thought, we have to react, open an issue, discuss about it and change it accordingly! 😉

Thanks again for your involvement!!

@devrck, I'm reviewing your PR right now then 🙂

composer.json
```diff
@@ -31,5 +31,16 @@
...
+    },
+    "scripts": {
+        "lint": [
+            "./vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --verbose --config=./.php_cs.dist --diff --show-progress=estimating --dry-run --using-cache=yes --allow-risky=yes"
+        ],
+        "lint:fix": [
+            "./vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --verbose --config=./.php_cs.dist --diff --show-progress=estimating --using-cache=no --allow-risky=yes"
+        ],
+        "test": [
+            "./vendor/bin/phpunit "
```
curquiza, line 43:
```suggestion
            "sh scripts/tests.sh"
```

src/Delegates/HandlesIndex.php
```diff
@@ -30,12 +38,15 @@ public function deleteAllIndexes(): void
         }
     }
 
-    public function getIndex($uid): Indexes
+    public function getIndex(string $uid): Indexes
```
curquiza, line 41: Since it has nothing to do with PHPCS, can you apply these kind of changes about type in another PR please? 🙂

src/Endpoints/Delegates/HandlesDocuments.php
```diff
@@ -1,12 +1,22 @@
...
+use MeiliSearch\Contracts\Http;
+use MeiliSearch\Exceptions\InvalidArgumentException;
+
+/**
+ * @property Http http
+ */
 trait HandlesDocuments
 {
-    public function getDocument(string $document_id)
+    public function getDocument($documentId)
```
curquiza, line 15: same here, except for the style naming

src/Endpoints/Delegates/HandlesDocuments.php
```diff
@@ -19,23 +29,35 @@ public function addDocuments(array $documents, string $primaryKey = null)
...
+        return $this->http->put(self::PATH.'/'.$this->uid.'/documents', $documents, ['primaryKey' => $primaryKey]);
     }
 
     public function deleteAllDocuments(): array
     {
         return $this->http->delete(self::PATH.'/'.$this->uid.'/documents');
     }
 
-    public function deleteDocument(string $document_id): array
+    public function deleteDocument($documentId): array
```
curquiza, line 42: Same here, except for the style naming

src/Exceptions/InvalidArgumentException.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
curquiza, line 1: Same

.php_cs.dist
```diff
@@ -1,12 +1,16 @@
 <?php
 
+declare(strict_types=1);
```
curquiza, line 3: Can you confirm this line has nothing to do with `native_function_invocation`? It's not a problem you can let it in the PR, I just want to be sure 🙂

src/Endpoints/Delegates/HandlesDocuments.php
```diff
@@ -19,23 +29,35 @@ public function addDocuments(array $documents, string $primaryKey = null)
...
+
+        return $this->http->delete(\sprintf('%s/%s/documents/%s', self::PATH, $this->uid, $documentId));
     }
 
     public function deleteDocuments(array $documents): array
     {
         return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete-batch', $documents);
     }
+
+    private function assertValidDocumentId($documentId): void
```
curquiza, line 54: This function seems to be really useful. But can you do it in another PR please?

src/Delegates/HandlesIndex.php
devrck: @curquiza 

Main reason for doing it was because of the `declare(strict_types=1);` It's good to use types in PHP but even better to enforce them with that directive.

When I applied the strict_type unit tests started breaking https://github.com/meilisearch/meilisearch-php/pull/56/commits/7ceea76528879fbc0552eba990939c9b13a4678e (https://github.com/meilisearch/meilisearch-php/runs/834289995) and that's why i removed the signature from here.

So maybe i can do a PR with applying just the simple native_invocation & a different PR for introducing `declare(strict_types=1);` but keep in mind that currently if somebody passed a `int` to getDocument it will fail.

.php_cs.dist
devrck: Yes, it has nothing to do with `native_function_invocation`, it's a way to tell PHP to enforce types.

src/Delegates/HandlesIndex.php
curquiza: Ok, I have'nt got that before, so I would rather have these changes in a new PR 🙂

Yes I keep in mind this is indeed a bug, I'm going to open an issue! Thanks @devrck for your explanation! 🚀

devrck: Alright will do the changes requested :wink: and we tackle this `declare(strict_types=1);` in a different PR.

src/Endpoints/Delegates/HandlesDocuments.php
devrck: Will be handled in https://github.com/meilisearch/meilisearch-php/issues/59

Issue #59, opened 2020-07-09
Apply declare_strict


curquiza:
Apply `declare_strict` in PHP CS, following [this discussion](https://github.com/meilisearch/meilisearch-php/pull/56#discussion_r452205214).

The problem is that currently, passing an int to `getDocument` makes it failed, but it should not since MS can accept int and string as document id.

@devrck feel free to complete this issue if you think it's not clear enough 🙂

Everyone is welcome to give their opinion about the subject here!

curquiza:
Closed with #62

PR #62, opened 2020-07-10
Apply declare_strict_types on codebase

devrck:
Fixing issue https://github.com/meilisearch/meilisearch-php/issues/59

### DIFF
--- a/src/Exceptions/InvalidArgumentException.php
+++ b/src/Exceptions/InvalidArgumentException.php
@@ -8,21 +8,21 @@ use Exception;
 
 class InvalidArgumentException extends Exception
 {
-    private $argumentName;
-
-    public function __construct(string $argumentName, array $validTypes)
+    public static function invalidType(string $argumentName, array $validTypes)
     {
-        $this->argumentName = $argumentName;
-
-        parent::__construct(
-            \sprintf('Argument "%s" is not a valid type! Please provide an argument that is of type: "%s"', $this->argumentName, \implode('","', $validTypes)),
+        return new self(
+            sprintf('Argument "%s" is not a valid type! Please provide an argument that is of type: "%s"', $argumentName, \implode('","', $validTypes)),
             400,
             null
         );
     }
 
-    public function getArgumentName(): string
+    public static function emptyArgument(string $argumentName)
     {
-        return $this->argumentName;
+        return new self(
+            sprintf('Argument "%s" is empty.', $argumentName),
+            400,
+            null
+        );
     }
 }
--- a/src/Exceptions/InvalidArgumentException.php
+++ b/src/Exceptions/InvalidArgumentException.php
@@ -8,7 +8,7 @@ use Exception;
 
 final class InvalidArgumentException extends Exception
 {
-    public static function invalidType(string $argumentName, array $validTypes)
+    public static function invalidType(string $argumentName, array $validTypes): self
     {
         return new self(
             sprintf('Argument "%s" is not a valid type! Please provide an argument that is of type: "%s"', $argumentName, \implode('","', $validTypes)),
@@ -17,7 +17,7 @@ final class InvalidArgumentException extends Exception
         );
     }
 
-    public static function emptyArgument(string $argumentName)
+    public static function emptyArgument(string $argumentName): self
     {
         return new self(
             sprintf('Argument "%s" is empty.', $argumentName),
--- a/src/Exceptions/InvalidArgumentException.php
+++ b/src/Exceptions/InvalidArgumentException.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MeiliSearch\Exceptions;
+
+use Exception;
+
+class InvalidArgumentException extends Exception
+{
+    private $argumentName;
+
+    public function __construct(string $argumentName, array $validTypes)
+    {
+        $this->argumentName = $argumentName;
+
+        parent::__construct(
+            \sprintf('Argument "%s" is not a valid type! Please provide an argument that is of type: "%s"', $this->argumentName, \implode('","', $validTypes)),
+            400,
+            null
+        );
+    }
+
+    public function getArgumentName(): string
+    {
+        return $this->argumentName;
+    }
+}
--- a/src/Exceptions/InvalidArgumentException.php
+++ b/src/Exceptions/InvalidArgumentException.php
@@ -6,7 +6,7 @@ namespace MeiliSearch\Exceptions;
 
 use Exception;
 
-class InvalidArgumentException extends Exception
+final class InvalidArgumentException extends Exception
 {
     public static function invalidType(string $argumentName, array $validTypes)
     {
--- a/tests/Endpoints/ClientTest.php
+++ b/tests/Endpoints/ClientTest.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Tests\Endpoints;
 
 use Http\Client\Exception\NetworkException;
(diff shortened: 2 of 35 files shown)


devrck:
@curquiza I find it strange why on my dev environment, following the steps of running tests it never is successful.

```bash
$ composer run-script test
> sh scripts/tests.sh
Setting FDs limit to 1000
Launching tests...
PHPUnit 8.5.8 by Sebastian Bergmann and contributors.

................................................................. 65 / 97 ( 67%)
..............F.................                                  97 / 97 (100%)

Time: 9.71 seconds, Memory: 8.00 MB

There was 1 failure:

1) Tests\Settings\DistinctAttributeTest::testUpdateDistinctAttribute
Failed asserting that 1 matches expected 'description'.

/home/devrck/projects/meilisearch-php/tests/Settings/DistinctAttributeTest.php:33

FAILURES!
Tests: 97, Assertions: 376, Failures: 1.
Script sh scripts/tests.sh handling the test event returned with error code 1

```

curquiza:
@devrck I tried to run the test your branch, it works 🙂

Did you do run:
```bash
$ docker pull getmeili/meilisearch:latest
```
?

Because it looks like the `latest` on machine is not the `latest` on Docker Hub 😉

devrck:
Yes that did the trick :smile: .

Btw why not wrap these commands into an composer run-script or makefile directive?

curquiza:
> Btw why not wrap these commands into an composer run-script or makefile directive?

Hum I'm not sure I want to apply transparently definitive changes on the Docker of our contributors... I would rather add an explained line in the contributing guideline so that they exactly know was is happening 🙂

tests/Settings/AcceptNewFieldsTest.php
```diff
@@ -8,9 +10,9 @@ class AcceptNewFieldsTest extends TestCase
 {
     private $index;
 
-    public function __construct()
+    public function setUp(): void
```
Guikingone, line 13: This method is actually `protected` in the parent class 🙂

tests/Settings/AttributesForFacetingTest.php
```diff
@@ -1,17 +1,13 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Tests\Settings;
 
 use Tests\TestCase;
 
 class AttributesForFacetingTest extends TestCase
```
Guikingone, line 9: Can be final?

tests/Settings/DistinctAttributeTest.php
```diff
@@ -8,9 +10,9 @@ class DistinctAttributeTest extends TestCase
 {
     private $index;
 
-    public function __construct()
+    public function setUp(): void
```
Guikingone, line 13: Same as before about visibility

tests/Settings/RankingRulesTest.php
```diff
@@ -17,9 +19,9 @@ class RankingRulesTest extends TestCase
         'exactness',
     ];
 
-    public function __construct()
+    public function setUp(): void
```
Guikingone, line 22: Same

tests/Settings/StopWordsTest.php
```diff
@@ -8,10 +10,9 @@ class StopWordsTest extends TestCase
 {
     private $index;
 
-    protected function setUp(): void
+    public function setUp(): void
```
Guikingone, line 13: Same

tests/TestCase.php
```diff
@@ -21,12 +23,19 @@ abstract class TestCase extends BaseTestCase
...
 
-    public function __construct()
+    public function setUp(): void
     {
-        parent::__construct();
+        parent::setUp();
         $this->client = new Client(self::HOST, self::DEFAULT_KEY);
+    }
+
+    public function tearDown(): void
```
Guikingone, line 37: That's also a `protected` method

src/Exceptions/InvalidArgumentException.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MeiliSearch\Exceptions;
+
+use Exception;
+
+class InvalidArgumentException extends Exception
```
Guikingone, line 9: Can be final

tests/Endpoints/ClientTest.php
```diff
@@ -10,12 +12,6 @@
 
 class ClientTest extends TestCase
 {
-    protected function setUp(): void
-    {
-        parent::setUp();
-        $this->client->deleteAllIndexes();
-    }
-
     public function testGetAllIndexesWhenEmpty()
```
Guikingone, line 15: Maybe we can add return typehint for each test method even if it's `: void`?

tests/Endpoints/DocumentsTest.php
```diff
@@ -1,8 +1,11 @@
...
 
+declare(strict_types=1);
+
 namespace Tests\Endpoints;
 
 use MeiliSearch\Exceptions\HTTPRequestException;
+use MeiliSearch\Exceptions\InvalidArgumentException;
 use Tests\TestCase;
 
 class DocumentsTest extends TestCase
```
Guikingone, line 11: Can be final

tests/Endpoints/ClientTest.php
devrck: This seems useless, I've added the `:void` to setUp because that's the method signature.

curquiza:
@devrck, sorry again, but could you [rebase your branch](https://gist.github.com/curquiza/5f7ce615f85331f083cd467fc4e19398), please? 🙂

I will merge just after that, since you seem having apply avery comment of @Guikingone.

src/Exceptions/InvalidArgumentException.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+namespace MeiliSearch\Exceptions;
+
+use Exception;
+
+final class InvalidArgumentException extends Exception
+{
+    public static function invalidType(string $argumentName, array $validTypes)
```
Guikingone, line 11: Rerurn typehint

src/Exceptions/InvalidArgumentException.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        return new self(
+            sprintf('Argument "%s" is not a valid type! Please provide an argument that is of type: "%s"', $argumentName, \implode('","', $validTypes)),
+            400,
+            null
+        );
+    }
+
+    public static function emptyArgument(string $argumentName)
```
Guikingone, line 20: Return typehint

Guikingone:
Hi 👋 

I've submitted 3 reviews for return typehint, should be good after that 🙂

devrck:
@Guikingone could you tell me what is the value we add by adding `:void` on test cases?

Guikingone:
@devrck The idea is to separate the tests of the data providers (which returns arrays or generator), it's easier to read and maintain if we know that every tests are executed without being dependent on other ones (which can occur if you don't specify the return typehint, we can be tempted to use this test as something else than just a test 🙂

devrck:
@Guikingone that make sense, thank you for the explanation, it didn't cross my mind :smile: .

tests/TestCase.php
```diff
@@ -21,16 +23,23 @@ abstract class TestCase extends BaseTestCase
 
     const DEFAULT_KEY = 'masterKey';
```
Guikingone, line 24: Consider adding a visibility for this const 🙂

devrck: Nice, added.

curquiza:
Thanks @devrck for your work and your patience! And thanks also @Guikingone for your review! Really helpful!

PR #64, opened 2020-07-10
Add explanation about how to fetch the MeiliSeach latest in CONTRIBUTING.md

curquiza:
I would like to add this line in the CONTRIBUTING.md file since a contributor was lost with an error in local that he could not see on the CI. See #62.

@eskombro and @bidoubiwa I would like to add this explanation in all the other SDKs, what do you think? 🙂


curquiza:
thanks @eskombro! Merging that! I will add it to all the other repositories soon

PR #82, opened 2020-08-29
add updateIndex method

shokme:


bidoubiwa:
Hey @shokme 
Thanks a lot for your contribution. @curquiza is on holliday but she'll be back soon enough to review your PR.
Meanwhile, do you think it is possible that you add some tests to check if your method works as expected?

shokme:
Hey, Indeed I forgot to add them !
While I was checking to add the test, I have seen a similar call. So I don't know if the new method is needed.

```php
    public function testPrimaryKeyUpdate(): void
    {
        $primaryKey = 'id';

        $response = $this->index->update(['primaryKey' => $primaryKey]);

        $this->assertSame($response['primaryKey'], $primaryKey);
        $this->assertSame('index', $response['uid']);
    }
```
But I also have seen that the two way exists for deleting an index.
using the `$client->deleteIndex($uid)` or `$index->delete()`

bidoubiwa (approved):

tests/Endpoints/IndexTest.php
```diff
@@ -141,6 +141,38 @@ public function testExceptionWhenPendingUpdateTimeOut(): void
...
+            'indexB',
+            ['primaryKey' => 'objectId']
+        );
+
+        $this->expectException(HTTPRequestException::class);
+
+        $this->client->updateIndex('indexB', ['primaryKey' => 'objectID']);
+    }
+
+    public function testExceptionIsThrownWhenUpdateIndexUseAnNoneExistingIndex(): void
```
eskombro, line 166:
```suggestion
    public function testExceptionIsThrownWhenUpdateIndexUseANoneExistingIndex(): void
```

eskombro (changes requested):
Hi @shokme!

Thanks for this contrib! 

It would be interesting to have this method if it was doing any additional check or simplifying access to the index update, but as you point out it is already accessible from an `$index->update(...)`, so I'm thinking it can indeed just create some unnecessary confusion to expose the same functionality in two different ways. So I am not sure if it is necessary to add it. Do you see any concrete advantage on adding it?

It would still be interesting to keep the tests in any case!

shokme:
> Do you see any concrete advantage on adding it?

To be honest no, I have made the pr because while writing the documentation I found the `deleteIndex($uid)` method without knowing the existence of `$index->update($options)`, so I was thinking the method was missing. ^^

We should keep it only if we want to reflect what we can already do with `delete`.
```php
$client->getIndex('foo')->delete();
// or
$client->deleteIndex('foo');
```

bidoubiwa:
I think both method are okey to keep if there is no code duplication. Both could be usefull as in some cases you'r wandering around with the `client` object and some cases with the `index` object.

But `$client->deleteIndex('foo');` should definitely call this function `$client->getIndex('foo')->delete();`

eskombro (approved):
Great! I agree! 👌

curquiza:
I'm okay with maintaining both methods doing the same thing: it's the purpose of an SDK, providing different methods for the different ways of coding 🙂
Obviously, if there is no code duplication as @bidoubiwa says!

Merging!!

PR #86, opened 2020-09-21
Fix locked in guzzle dependency

ppshobi:
Follow up PR for #81, with updated readme to install guzzle-6 adapter


curquiza (changes requested):
`symfony/http-client`, `php-http/curl-client` and `kriswallsmith/buzz` need `nyholm/psr7` to work 🙂
Could you update the README please?

curquiza (approved):

PR #87, opened 2020-09-21
Update docs for HTTP Client compatibilities

curquiza:
Introduction sentence based on https://github.com/GitLabPHP/Client#installation

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ To get started, simply require the project using [Composer](https://getcomposer.
 You will also need to install packages that "provide" [`psr/http-client-implementation`](https://packagist.org/providers/psr/http-client-implementation) and [`psr/http-factory-implementation`](https://packagist.org/providers/psr/http-factory-implementation).<br>
 A list with compatible HTTP clients and client adapters can be found at [php-http.org](http://docs.php-http.org/en/latest/clients.html).
 
-**If you don't know what HTTP client to require, we recommend using Guzzle** with the `php-http/guzzle6-adapter`:
+**If you don't know which HTTP client to require, we recommend using Guzzle** with the `php-http/guzzle6-adapter`:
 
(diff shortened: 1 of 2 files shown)


ppshobi:
Looking good for me :)

README.md
```diff
@@ -34,29 +34,28 @@
...
-with the package to avoid package collisions with existing libraries or projects. A list with compatible HTTP clients and client adapters can be found at [php-http.org](http://docs.php-http.org/en/latest/clients.html).
 
-### Most users probably want the following.
-Using composer:
+To get started, simply require the project using [Composer](https://getcomposer.org/).<br>
+You will also need to install packages that "provide" [`psr/http-client-implementation`](https://packagist.org/providers/psr/http-client-implementation) and [`psr/http-factory-implementation`](https://packagist.org/providers/psr/http-factory-implementation).<br>
+A list with compatible HTTP clients and client adapters can be found at [php-http.org](http://docs.php-http.org/en/latest/clients.html).
 
-With `guzzle6-adapter`
+**If you don't know which HTTP client to require, we recommend using Guzzle** with the `php-http/guzzle6-adapter`:
```
bidoubiwa, line 46:
```suggestion
**If you don't know which HTTP client to use, we recommend using Guzzle** with the `php-http/guzzle6-adapter`:
```

bidoubiwa (changes requested):

README.md
curquiza: I use require because that's the term used when using `composer`:
```bash
$ composer require your-package
```

bidoubiwa: I think out of strict composed context this sentence is confusing as `require` implies a condition and not a fetch. I have not changed it in the previous paragraph since it was in a composer context. In this sentence, add works both if you have or not the composer context.

PR #98, opened 2020-09-28
fix HTTPRequestException when body is null

shokme:
sometimes the body response is `null` and even if `null` looks like authorized it will throw this error:

```bash
Error : Wrong parameters for MeiliSearch\Exceptions\HTTPRequestException([string $message [, long $code [, Throwable $previous = NULL]]])
 .../meilisearch-php/src/Exceptions/HTTPRequestException.php:21
```

by returning `$response->getReasonPhrase()` the error should not happen again

```php
....
* @return string Reason phrase; must return an empty string if none present.
*/
public function getReasonPhrase();
```
In my test case:
using a wrong url -> http://localhost instead of http://localhost:7700 will return

```bash
MeiliSearch\Exceptions\HTTPRequestException : Not Found
.../meilisearch/meilisearch-php/src/Http/Client.php:183
```

I guess this will also solve https://github.com/meilisearch/meilisearch-laravel-scout/issues/37

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -178,7 +178,7 @@ class Client implements Http
     private function parseResponse(ResponseInterface $response)
     {
         if ($response->getStatusCode() >= 300) {
-            $body = json_decode($response->getBody()->getContents(), true);
+            $body = json_decode($response->getBody()->getContents(), true) ?? $response->getReasonPhrase();
             throw new HTTPRequestException($response->getStatusCode(), $body);
         }
 


curquiza:
Hello @shokme!
Thanks for this PR, it seems great 🙂

A question to be sure I understand your fix:

This example
```
<?php

require_once __DIR__ . '/vendor/autoload.php';
use MeiliSearch\Client;
$client = new MeiliSearch\Client('http://localhost:770/', 'masterKey');
$index = $client->version();
```
currently fails (`v0.14.1`) with this error:
```bash
$ php test.php
Fatal error: Uncaught Buzz\Exception\NetworkException: file_get_contents(http://localhost//version): failed to open stream: Connection refused in /Users/curquiza/Documents/tests/php2/vendor/kriswallsmith/buzz/lib/Client/FileGetContents.php:26
Stack trace:
#0 /Users/curquiza/Documents/tests/php2/vendor/meilisearch/meilisearch-php/src/Http/Client.php(165): Buzz\Client\FileGetContents->sendRequest(Object(Nyholm\Psr7\Request))
#1 /Users/curquiza/Documents/tests/php2/vendor/meilisearch/meilisearch-php/src/Http/Client.php(81): MeiliSearch\Http\Client->execute(Object(Nyholm\Psr7\Request))
#2 /Users/curquiza/Documents/tests/php2/vendor/meilisearch/meilisearch-php/src/Contracts/Endpoint.php(21): MeiliSearch\Http\Client->get('/version')
#3 /Users/curquiza/Documents/tests/php2/vendor/meilisearch/meilisearch-php/src/Delegates/HandlesSystem.php(29): MeiliSearch\Contracts\Endpoint->show()
#4 /Users/curquiza/Documents/tests/php2/test.php(11): MeiliSearch\Client->version()
#5 {main}
  thrown in /Users/curquiza/Documents/tests/php2/vendor/kriswallsmith/buzz/lib/Client/FileGetContents.php on line 26
```

With your fix, and the same example, it would return:

```bash
MeiliSearch\Exceptions\HTTPRequestException : Not Found
.../meilisearch/meilisearch-php/src/Http/Client.php:183
```

Do I understand well?
If I'm right, could you update this test to be more accurate? 🙂
https://github.com/meilisearch/meilisearch-php/blob/5c0562c4c35bc4364c4b14f48157e25da5fc5f3a/tests/Endpoints/ClientTest.php#L253-L264

shokme:
Hey !

No this is very different, here your exception came before getting the response. this is thrown by your HTTPClient (sendRequest() method) and the message said "uncaught" so this exception not handled.

if you want to handle it you can do something like this:

```php
// from the execute() method inside src/Http/Client.php
try {
    return $this->parseResponse($this->http->sendRequest($request));
} catch (NetworkExceptionInterface $e) {
    throw new HTTPRequestException($e->getCode(), $e->getMessage(), $e->getPrevious());
}
```
If you want to handle it, I can add it to this PR or another as you wish.

My PR fix the response exception, in my case http://localhost is a simple nginx. that's why I don't encounter the HTTPClient "Couldn't connect ..." and if I get localhost/version it will return a 404 response.

curquiza:
Ok! I didn't understand it was to handle nginx errors! Good 🙂

> If you want to handle it, I can add it to this PR or another as you wish.

No, it's ok, I just wanted to understand, this behavior is not an emergency to handle for the moment 😉

Great! Going to merge it!

curquiza (approved):
🚀

PR #100, opened 2020-09-28
Remove the ability to replace Response and Request Factories

### LINKED ISSUE #89: Removing the ability to replace the HTTP Client
The replacing client part needs to be removed since our SDK should automatically discover the installed client.

Linked issue: #76

ppshobi:
closes #89

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@
   - [Update status](#update-status)
   - [Search](#search)
 - [🧰 HTTP Client Compatibilities](#-http-client-compatibilities)
+    - [Replacing Http Client](#-replacing-http-client)
 - [⚙️ Development Workflow and Contributing](#️-development-workflow-and-contributing)
 
 ## 🔧 Installation
@@ -326,7 +327,18 @@ $ composer require meilisearch/meilisearch-php php-http/curl-client nyholm/psr7:
 ```bash
 $ composer require meilisearch/meilisearch-php kriswallsmith/buzz nyholm/psr7:^1.0
 ```
+### Replacing http client
+For some reason you want to pass in your own http client with a custom configuration, you could make use of this feature. Make sure you have a psr 18 compatible client when you initialize the Meilisearch client. 
 
+A good use case can be found [here](https://github.com/meilisearch/meilisearch-php/pull/100#issuecomment-700577627)
+
+Pass in your client as the third argument to the `Meilisearch/Client` constructor 
+
+eg:
+
+```php
+new Client(MEILISEARCH_ENDPOINT, MEILISEARCH_API_KEY, new GuzzleHttpClient(['timeout' => 2]));
+```
 ## ⚙️ Development Workflow and Contributing
 
 Any new contribution is more than welcome in this project!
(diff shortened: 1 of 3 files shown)


bensherred (approved):

curquiza (dismissed):
LGTM 🙂
Thanks!!

(waiting for @shokme approval if he's available 😉)

fab120:
I do not think that removing this feature is a good choice.

Currently, replacing the http client is the only way to initialize it with a custom configuration.

For example in my use case:
```php
        $this->app->singleton(Client::class, function (Application $app) {
            return new Client(
                config('meilisearch.url'),
                config('meilisearch.apikey'),
                new GuzzleHttpClient(['timeout' => 2]),
            );
        });
```

Guzzle comes out with "wait indefinitely" behavior by default and if there is any connection problem, a request to Meilisearch tooks at least php `max_execution_time` before fail.

shokme:
AS @fab120 said, keeping the client configuration is better.

But **I think**, we can remove the Request and StreamFactory as they don't look like to have a constructor like an http client.

ppshobi:
This is something specific to guzzle I believe, so keeping that code for something specific to a library defeats the prospect of a client agnostic library like this. Later on, we will be forced to add additional functionalities to satisfy specific client library needs. 

But at the same time, I don't want to remove this flexibility as well :|

fab120:
@ppshobi Looking at the documentation of clients suggested [here](https://github.com/meilisearch/meilisearch-php#-http-client-compatibilities), I can see that everyone can be initialized with some configuration: [symfony/http-client](https://symfony.com/doc/current/http_client.html#configuration), [php-http/curl-client](http://docs.php-http.org/en/latest/clients/curl-client.html#configuring-client), [kriswallsmith/buzz](https://github.com/kriswallsmith/Buzz/blob/master/doc/client.md#configuration).

So I would not say this is "something specific to guzzle".
Any user can have any reason to initialize their chosen client with options suitable for their use case.

I gave an example specific to my chosen client (guzzle) and my use case where if a response takes more than 2s means there is something I have to look into.

curquiza:
Hello guys!
Indeed we might let this configuration: it does not bring a lot of code complexity (for the moment) and add a nice feature as @fab120 explains 🙂

And what do you think @ppshobi and @fab120 about the @shokme suggestion?

> But I think, we can remove the Request and StreamFactory as they don't look like to have a constructor like an http client.

ppshobi:
Yes, I have updated the PR. Also wrote a readme section regarding the same.

fab120:
> And what do you think @ppshobi and @fab120 about the @shokme suggestion?

I agree with @shokme 

@ppshobi The pr looks good to me.
If you want to keep the readme more agnostic, I think something like this could be better:
```php
 new Client(MEILISEARCH_ENDPOINT, MEILISEARCH_API_KEY, new HttpClient($config));
```

shokme (approved):

curquiza (approved):
LGTM!
@fab120 I will slightly update the README with the person in charge of the docs in Meili so that @ppshobi does not bother with this kind of details 😉

curquiza:
Thanks @ppshobi for this PR 🚀
So nice to you @fab120 to watch this repository, your remark was really helpful!

PR #104, opened 2020-09-30
Update delete index to always return an array

### LINKED ISSUE #79: Delete index method should return array
In the endpoint class `Indexes`, the method `delete()` seems to return `void`:
https://github.com/meilisearch/meilisearch-php/blob/dc506d60be544b996558f64d20d7c70ca8278810/src/Endpoints/Indexes.php#L80

However, the HTTP `Client` method `delete` seems to return the value of the executed query:
https://github.com/meilisearch/meilisearch-php/blob/dc506d60be544b996558f64d20d7c70ca8278810/src/Http/Client.php#L150

Related to Symfony bundle too, the bundle will probably return an error when trying to delete an entry, because is seems to return an `array` in the next line: 
https://github.com/meilisearch/meilisearch-symfony/blob/3714b3cbb1a5c9a7e8a57b289e197f548b532cc7/src/Engine.php#L140

So we need to decide if we return `void` from the beginning to the end or an `array`.   
From my perspective, I would prefer to return an `array` with the response if the index has been deleted correctly or not.

pedroxido:
Fixes #79
As per the discussion in the related issue, the method `HandlesIndex::deleteIndex` returns void but should return an array given that Meilisearch delete api could return an array in the future.

Added few lines to always return the array from the Meilisearch delete response, if null, returns an empty array

### DIFF
--- a/src/Delegates/HandlesIndex.php
+++ b/src/Delegates/HandlesIndex.php
@@ -25,9 +25,9 @@ trait HandlesIndex
         return (new Indexes($this->http, $uid))->show();
     }
 
-    public function deleteIndex(string $uid): void
+    public function deleteIndex(string $uid): array
     {
-        (new Indexes($this->http, $uid))->delete();
+        return (new Indexes($this->http, $uid))->delete();
     }
 
     public function deleteAllIndexes(): void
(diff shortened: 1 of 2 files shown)


emulienfou:
@curquiza Indeed, perfect for me

curquiza (approved):
Thanks @pedroxido for your work!

And thanks @ppshobi and @emulienfou for your incredible reactivity, as usual 😁

PR #124, opened 2020-12-30
 Changes due to the implicit index creation

curquiza:
Related to https://github.com/meilisearch/integration-guides/issues/48

Changes:
- New method `$client->index($uid)`: replaces `$client->getIndex($uid)`. Does not do any HTTP call but only returns an `Index` instance.
- New attribute in the `Index` class: `primaryKey`
- New method `$index->fetchRawInfo()`: returns an array of index
- New method `$index->fetchInfo()`: returns an `Index` instance

Breaking changes:
- `$client->getIndex($uid)` now does an HTTP call. Use `$client->index($uid)` instead.
- `$client->updateIndex` returns an `Index` instance.
- `$index->getPrimaryKey()` does not do any HTTP call, but only return the `primaryKey` attribute. Use `$index->fetchPrimaryKey()` if you want the same behavior as the old `$index->getPrimaryKey()`
- `$index->show()` is removed and replaced by `$index->fetchInfo()` (returns an instance) and `$index->fetchRawInfo()` (returns an array)
- `$client->showIndex($uid)` is removed. Use `$client->getIndex($uid)` (returns an instance) or `$client->index($uid)->fetchRawInfo()` (returns an array) instead.

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -4,7 +4,7 @@
 # You can read more on https://github.com/meilisearch/documentation/tree/master/.vuepress/code-samples
 ---
 get_one_index_1: |-
-  $client->index('movies')->getRaw();
+  $client->index('movies')->fetchRawInfo();
 list_all_indexes_1: |-
   $client->getAllIndexes();
 create_an_index_1: |-
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -4,7 +4,7 @@
 # You can read more on https://github.com/meilisearch/documentation/tree/master/.vuepress/code-samples
 ---
 get_one_index_1: |-
-  $client->getIndex('movies');
+  $client->index('movies')->getRaw();
 list_all_indexes_1: |-
   $client->getAllIndexes();
 create_an_index_1: |-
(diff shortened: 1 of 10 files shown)


.code-samples.meilisearch.yaml
```diff
@@ -4,25 +4,25 @@
 # You can read more on https://github.com/meilisearch/documentation/tree/master/.vuepress/code-samples
 ---
 get_one_index_1: |-
-  $client->getIndex('movies');
+  $client->index('movies')->getRaw();
```
eskombro, line 7: Shouldn't this use `fetchInfo()` instead of `getRaw()`, returning an Index instead of an http response?

eskombro (approved):
Great implementation @curquiza ! TY

🎉

.code-samples.meilisearch.yaml
curquiza: we chose to let the raw option for the moment -> https://github.com/meilisearch/meilisearch-js/pull/715/files#r539555931

eskombro: Ok, then everything LGTM 🎉

eskombro (approved):
👌🏽

curquiza:
Bors merge

PR #134, opened 2021-01-20
Improve the $options parameter in search() method

curquiza:
- rename `toJson` into `toJSON`: same naming as Stripe and Twilio
- add `rawSearch()` method
- `transformHits` callback can take any callback to completely transform the hits
- Add `transformFacetsDistribution` option in `search()` method (type: callback)
- Add `removeZeroFacets` option in `search()` method (type: boolean)
- Add `transformHits()`, `transformFacetsDistribution()` and `removeZeroFacets()` methods.

### Usage

```php
$searchResult = $index->search('prince');
$searchResult->removeZeroFacets()

// OR

$searchResult = $index->search('prince', [], ['removeZeroFacets' => true]);
```

```php
$keepAmericanSniperFunc = function (array $hits) {
    return array_filter(
        $hits,
        function (array $hit) { return 'American Sniper' === $hit['title']; }
    );
};

$searchResult = $index->search('american');
$searchResult->transformHits($keepAmericanSniperFunc)

// OR

$response = $this->index->search('prince', [], $options = ['transformHits' => $keepAmericanSniperFunc]);
```

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -46,10 +46,14 @@ delete_one_document_1: |-
 delete_documents_1: |-
   $client->index('movies')->deleteDocuments([23488, 153738, 437035, 363869]);
 search_1: |-
-  $searchResults = $client->index('movies')->search('american ninja')
-  $searchResults->getHits();
-  $searchResults->getFacetsDistribution();
-  $searchResults->getRaw(); // The JSON response returned by MeiliSearch
+  // Do a search
+  $searchResults = $client->index('movies')->search('american ninja');
+
+  // Get results in an Array using a getter
+  $hits = $searchResults->getHits();
+
+  // Get the decoded response of MeiliSearch, see response below
+  $response = $searchResults->getRaw();
 get_update_1: |-
   $client->index('movies')->getUpdateStatus(1);
 get_all_updates_1: |-
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -46,7 +46,10 @@ delete_one_document_1: |-
 delete_documents_1: |-
   $client->index('movies')->deleteDocuments([23488, 153738, 437035, 363869]);
 search_1: |-
-  $client->index('movies')->search('american ninja');
+  $searchResults = $client->index('movies')->search('american ninja')
+  $searchResults->getHits();
+  $searchResults->getFacetsDistribution();
+  $searchResults->getRaw(); // The JSON response returned by MeiliSearch
 get_update_1: |-
   $client->index('movies')->getUpdateStatus(1);
 get_all_updates_1: |-
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ With the `updateId`, you can check the status (`enqueued`, `processed` or `faile
 
 ```php
 // MeiliSearch is typo-tolerant:
-print_r($index->search('harry pottre'));
+print_r($index->search('harry pottre')->getHits());
 ```
 
 Output:
@@ -118,20 +118,11 @@ Output:
 ```php
 Array
 (
-    [hits] => Array
+    [0] => Array
         (
-            [0] => Array
-                (
-                    [id] => 4
-                    [title] => Harry Potter and the Half-Blood Prince
-                )
-
+            [id] => 4
+            [title] => Harry Potter and the Half-Blood Prince
         )
-
-    [offset] => 0
-    [limit] => 20
-    [processingTimeMs] => 1
-    [query] => harry pottre
 )
 ```
 
@@ -139,14 +130,19 @@ Array
 
 All the supported options are described in the [search parameters](https://docs.meilisearch.com/guides/advanced_guides/search_parameters.html) section of the documentation.
 
+💡 **More about the `search()` method in [the Wiki](https://github.com/meilisearch/meilisearch-php/wiki/Search).**
+
 ```php
-$index->search('prince',
+$index->search(
+    'prince',
     [
         'attributesToHighlight' => ['*'],
         'filters' => 'book_id > 10'
     ]
-);
+)->getRaw();
 ```
+
+JSON output:
 
 ```json
 {
@@ -161,15 +157,6 @@ $index->search('prince',
     "processingTimeMs": 10,
     "query": "prince"
 }
-```
-
-With `filters`, both single and double quotes are supported.
-```php
-// Enclosing with double quotes
-$index->search('prince', ['filters' => "title = 'Le Petit Prince' OR author = 'J. R. R. Tolkien'"]);
-
-// Enclosing with single quotes
-$index->search('hobbit', ['filters' => 'title = "The Hitchhiker\'s Guide to the Galaxy" OR author = "J. R. R. Tolkien"']);
 ```
 
 ## 🤖 Compatibility with MeiliSearch
--- a/README.md
+++ b/README.md
@@ -110,7 +110,8 @@ With the `updateId`, you can check the status (`enqueued`, `processed` or `faile
 
 ```php
 // MeiliSearch is typo-tolerant:
-print_r($index->search('harry pottre')->getHits());
+$hits = $index->search('harry pottre')->getHits();
+print_r($hits);
 ```
 
 Output:
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ $index->search(
         'attributesToHighlight' => ['*'],
         'filters' => 'book_id > 10'
     ]
-)->getRaw();
+)->getRaw(); // Return in Array format
 ```
 
 JSON output:
--- a/src/Endpoints/Indexes.php
+++ b/src/Endpoints/Indexes.php
@@ -152,7 +152,7 @@ class Indexes extends Endpoint
         }
 
         $searchResult = new SearchResult($result);
-        $searchResult = $searchResult->applyOptions($options);
+        $searchResult->applyOptions($options);
 
         return $searchResult;
     }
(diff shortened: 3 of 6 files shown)


curquiza:
Please @bidoubiwa, take also a look at the WIP wiki which is tightly linked to this PR -> https://github.com/meilisearch/meilisearch-php/wiki/Search

README.md
```diff
@@ -110,44 +110,40 @@ With the `updateId`, you can check the status (`enqueued`, `processed` or `faile
 
 ```php
 // MeiliSearch is typo-tolerant:
-print_r($index->search('harry pottre'));
+print_r($index->search('harry pottre')->getHits());
```
bidoubiwa, line 113: What would the complete instance look like? Without getHits ?

src/Endpoints/Indexes.php
```diff
@@ -145,26 +145,33 @@ public function waitForPendingUpdate($updateId, $timeoutInMs = 5000, $intervalIn
...
         if (\array_key_exists('raw', $options) && $options['raw']) {
             return $result;
         }
 
         $searchResult = new SearchResult($result);
-
-        if (\array_key_exists('transformHits', $options) && \is_callable($options['transformHits'])) {
-            $searchResult = $searchResult->filter($options['transformHits']);
-        }
+        $searchResult = $searchResult->applyOptions($options);
```
bidoubiwa, line 155: If `applyOptions` is always used after `new SearchResult`, should it not be in the constructor of `SearchResult` ?

src/Search/SearchResult.php
```diff
@@ -84,16 +94,53 @@ public function __construct(array $body)
      *
      * @return SearchResult
      */
```
bidoubiwa, line 96: Is this not outdated?
```
/**
     * Return a new {@see SearchResult} instance with the hits filtered using `array_filter($this->hits, $callback, ARRAY_FILTER_USE_BOTH)`.
     *
     * The $callback receives both the current hit and the key, in that order.
     *
     * The method DOES not trigger a new search.
     *
     * @return SearchResult
     */
```

As per [Wikipedia](https://en.wikipedia.org/wiki/Fluent_interface#PHP): 
> In PHP, one can return the **current object** by using the `$this` special variable which represent the instance. Hence return $this; will make the method return the instance. The example below defines a class Employee and three methods to set its name, surname and salary. Each return the instance of the Employee class allowing to chain methods.


> Return a new {@see SearchResult} instance with the hits filtered using...

Should it not be `Return the current instance updated with the hits filtered using...

( I cannot suggest changed as it was not modified in this PR )

>      * The $callback receives both the current hit and the key, in that order.

Only some of the options require a callback.

I think unless I understand badly in which case sorry, it was not updated with the change from filter to applyOptions.

src/Endpoints/Indexes.php
bidoubiwa: Also, are modifications not apply directly on the instance `SearchResult`? I'm not sure ` $searchResult =` is needed.

src/Search/SearchResult.php
```diff
@@ -84,16 +94,53 @@ public function __construct(array $body)
...
+                    $facet,
+                    function ($v, $k) { return 0 !== $v; },
+                    ARRAY_FILTER_USE_BOTH
+                );
+            };
+
+            return array_map($filterOneFacet, $facets);
+        };
+
+        return $this->transformFacetsDistribution($filterAllFacets);
```
bidoubiwa, line 141: Just curiosity
Would this be a possible or better alternative?

```php
$this->facetsDistribution = $filterAllFacets($facetsDistribution);

return $this
```

or even this: 

```php

 public function removeZeroFacets(): self
    {
        $filterAllFacets = function (array $facets) {
            $filterOneFacet = function (array $facet) {
                return array_filter(
                    $facet,
                    function ($v, $k) { return 0 !== $v; },
                    ARRAY_FILTER_USE_BOTH
                );
            };

            $this->transformFacetsDistribution = array_map($filterOneFacet, $facets);
        };

        $filterAllFacets($this->facetsDistribution);

        return $this;
}
```

In both cases `transformFacetsDistribution` wouldn't be useful. But If you want to use the method you created I understand. But I think in either cases you can return `$this` no ? instead of 
```php
return $this->transformFacetsDistribution($filterAllFacets);
```
which result is not used

src/Search/SearchResult.php
```diff
@@ -119,12 +166,17 @@ public function getLimit(): int
 
     public function getHitsCount(): int
     {
-        return $this->nbHits;
+        return $this->hitsCount;
+    }
+
+    public function count(): int
+    {
+        return $this->hitsCount;
```
bidoubiwa, line 174: There are functions with comments and some without. I'm not sure this is important right now but it might impact IntelliSense's relevancy.

bidoubiwa (changes requested):

README.md
curquiza: It's an instance, it does not look anything readable, it's too long.
And in some languages, an instance is not even displayable without using a method in the class.

src/Search/SearchResult.php
curquiza: Yes, IntelliSence is indeed a future project, but not a priority for the MVP here.

README.md
```diff
@@ -110,44 +110,40 @@ With the `updateId`, you can check the status (`enqueued`, `processed` or `faile
...
 ```php
-$index->search('prince',
+$index->search(
+    'prince',
     [
         'attributesToHighlight' => ['*'],
         'filters' => 'book_id > 10'
     ]
-);
+)->getRaw();
```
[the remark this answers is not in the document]
curquiza: The instance is not friendly at all to display.
Either we show the user there are `getHits()` and `getFacetsDistribution()` (the most used ones) or we show the way to get the complete answer with `getRaw()` or `toArray()`

src/Endpoints/Indexes.php
curquiza: I'm not sure I want to pass it in the constructor. It's not because you create a `SearchResult()` you want automatically the options to be applied.
This class is public so our users can use it in their own ways.
I agree, catch the return is not mandatory.

src/Search/SearchResult.php
curquiza: Oh! Thanks! I forgot to update the comment!

src/Search/SearchResult.php
curquiza: I would rather use the `transformFacetsDistribution` function.
Indeed, if at some point, there is another field to update when transforming the `facetsDistribution` (as it is for `transformHits` where we update the `hits` AND the `hitsCount`), I would rather all be done in the same function to avoid duplication, so to avoid mistakes.

.code-samples.meilisearch.yaml
```diff
@@ -46,7 +46,10 @@ delete_one_document_1: |-
 delete_documents_1: |-
   $client->index('movies')->deleteDocuments([23488, 153738, 437035, 363869]);
 search_1: |-
-  $client->index('movies')->search('american ninja');
+  $searchResults = $client->index('movies')->search('american ninja')
+  $searchResults->getHits();
+  $searchResults->getFacetsDistribution();
+  $searchResults->getRaw(); // The JSON response returned by MeiliSearch
```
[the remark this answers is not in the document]
curquiza: > But in my humble opinion, in a very simple approach, I would keep the following line and not the raw one as per default PHP users will expect a PHP class instance.

>  $searchResults = $client->index('movies')->search('american ninja')

The problem with this: the user does not even know what are the main methods exist.

We cannot follow algolia or elastic -> they don't return an instance class like us. They provide "simple" wrapper.

I did this suggestion because I don't want our user to be lost with a huge class he/she can't even read when trying to display it.
But I really open to improvement, I felt a little bit confused about what to do here.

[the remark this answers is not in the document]
bidoubiwa: Mmmh, if we had to add it to the search API references I think we should adapt every SDK to showcase the same possibilities. 

Also, if we add your example, I think context is needed. For example:

```php
   // Do a search
   $searchResults = $client->index('movies')->search('american ninja')

  // Get results in array
  $hits = $searchResults->getHits();
  $facetDistribution = $searchResults->getFacetsDistribution();

  // Get results in JSON
  $searchResults->getRaw(); // JSON encoded answer of MeiliSearch, see response bellow
```

src/Endpoints/Indexes.php
bidoubiwa:
```suggestion
      $searchResult->applyOptions($options);
```

src/Search/SearchResult.php
bidoubiwa: And about changing this to  `  return $this->transformFacetsDistribution($filterAllFacets);`

```php
    $this->transformFacetsDistribution($filterAllFacets);

    return $this;
```

which would avoid making this return the facetDistribution and thus changing the value of searchResponse.
```php
$searchResponse = $this-removeZeroFacets() 
```


This would avoid having to do this:

curquiza: If I'm not wrong, maybe I miss-understand something, but here, your alternative and my original suggestion are exactly the same

.code-samples.meilisearch.yaml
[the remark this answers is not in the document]
curquiza:
```php
   // Do a search
   $searchResults = $client->index('movies')->search('american ninja')

  // Get results in an Array using a getter
  $hits = $searchResults->getHits();

  // Get the decoded response of MeiliSearch, see response below
  $response = $searchResults->getRaw();
```

README.md
curquiza, line 113:
```suggestion
$hits = $index->search('harry pottre')->getHits();
print_r($hits);
```

README.md
curquiza, line 142:
```suggestion
)->getRaw(); // Return in Array format
```

bidoubiwa (approved):
LGTM 🔥 Great work

curquiza:
bors merge

PR #146, opened 2021-02-22
Should throw exception if httpBody empty

jwohlfert23:
I've been seeing the occasional error in my app saying "index already exists". I traced that error back to the "getOrCreateIndex" method.

When the response from meilisearch is empty, it should not try to create the index because we don't know for sure that the index "not found".

### DIFF
--- a/src/Delegates/HandlesIndex.php
+++ b/src/Delegates/HandlesIndex.php
@@ -61,10 +61,11 @@ trait HandlesIndex
         try {
             $index = $this->getIndex($uid, $options);
         } catch (HTTPRequestException $e) {
-            if (\is_array($e->httpBody) && 'index_not_found' !== $e->httpBody['errorCode']) {
+            if (\is_array($e->httpBody) && 'index_not_found' === $e->httpBody['errorCode']) {
+                $index = $this->createIndex($uid, $options);
+            } else {
                 throw $e;
             }
-            $index = $this->createIndex($uid, $options);
         }
 
         return $index;


bidoubiwa:
Hey @jwohlfert23. Thanks a lot for this contribution. If I understand correctly. In the case another error occurs, that is not `not_found`, there should not be an attempt to create a new index? 

I think this is a great addition. Do you have an example of when or how this problem occurred?

jwohlfert23:
In my case, the issue is sometimes that the service is unavailable, meaning I'm getting a 400 or 500 level error, with no response body.  When there is an empty body, it is trying to create the index, even though the error isn't specifying that explicitly.

src/Delegates/HandlesIndex.php
```diff
@@ -61,7 +61,7 @@ public function getOrCreateIndex(string $uid, array $options = []): Indexes
         try {
             $index = $this->getIndex($uid, $options);
         } catch (HTTPRequestException $e) {
-            if (\is_array($e->httpBody) && 'index_not_found' !== $e->httpBody['errorCode']) {
+            if (!\is_array($e->httpBody) || 'index_not_found' !== $e->httpBody['errorCode']) {
                 throw $e;
             }
             $index = $this->createIndex($uid, $options);
```
bidoubiwa, line 67:
```suggestion
            if (\is_array($e->httpBody) && 'index_not_found' == $e->httpBody['errorCode'])
            {
                $index = $this->createIndex($uid, $options);
            } else {
               throw $e;
            }
```

The first part of the if the condition was probably added to ensure the `httpBody` has content in it. As it could be an error not related to MeiliSearch in which case the httpBody might be empty.

I try to find a more bulletproof solution. What do you think of this suggestion?

bidoubiwa (changes requested):

src/Delegates/HandlesIndex.php
jwohlfert23: works for me!

bidoubiwa: Could you commit the suggestion?

jwohlfert23: Done

bidoubiwa (approved):

bidoubiwa:
bors merge

PR #151, opened 2021-03-22
Add .idea in gitignore

claudiunicolaa:
If using one of the JetBrains IDEA, every time you add a file to a commit you need to pay attention to do not add the `.idea` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 composer.lock
 /vendor/
 .php_cs.cache
+/.idea/


shokme:
Hello @claudiunicolaa,

this is better to add it to your global .gitignore config.

example: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

bidoubiwa:
Hello @claudiunicolaa
Thanks a lot for taking the time to contribute. Following @shokme suggestion, I will close this issue but thanks anyway!

PR #155, opened 2021-03-29
Separate tests that ignored second test

bidoubiwa:
In these tests since the first test throws an error the second is ignored. Separating them to test uid empty string case.


curquiza (approved):
Top!
Merge when you want :)

bidoubiwa:
bors merge

PR #157, opened 2021-04-15
Standardize health method

alallema:
Checking that method health() return `{'status': 'available'}`  and added isHealthy() method who return boolean value

meilisearch/integration-guides#55

### DIFF
--- a/src/Delegates/HandlesSystem.php
+++ b/src/Delegates/HandlesSystem.php
@@ -24,6 +24,17 @@ trait HandlesSystem
         return $this->health->show();
     }
 
+    public function isHealthy(): bool
+    {
+        try {
+            $this->health->show();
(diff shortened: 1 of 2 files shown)


src/Delegates/HandlesSystem.php
```diff
@@ -24,6 +24,17 @@ public function health(): ?array
         return $this->health->show();
     }
 
+    public function isHealthy(): bool
+    {
+        try {
+            $health = $this->health->show();
```
bidoubiwa, line 30:
```suggestion
            $this->health->show();
```
Not sure you need to use a variable.

bidoubiwa (commented):
Just a small suggestion but apart from that LGTM 🔥

bidoubiwa (approved):
bors merge

PR #159, opened 2021-04-23
Add explicit types to parameters

mmachatschek:
These routes receive arrays (as of the documention) as parameters and therefore should enforce the array type to these specific routes.

### DIFF
--- a/src/Endpoints/Delegates/HandlesSettings.php
+++ b/src/Endpoints/Delegates/HandlesSettings.php
@@ -52,7 +52,7 @@ trait HandlesSettings
         return $this->http->get(self::PATH.'/'.$this->uid.'/settings/searchable-attributes');
     }
 
-    public function updateSearchableAttributes($searchableAttributes): array
+    public function updateSearchableAttributes(array $searchableAttributes): array
     {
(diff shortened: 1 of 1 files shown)


curquiza (approved):
LGTM, thanks a lot @mmachatschek! :)

We wait for the review of @bidoubiwa before any merge 🙂

bidoubiwa (approved):
Hello, thanks a lot for this PR! 🔥

I have a question. I noticed in other files that typing was not always done. For example
```php
   public function getDocument($documentId)
    {
        $this->assertValidDocumentId($documentId);

        return $this->http->get(self::PATH.'/'.$this->uid.'/documents/'.$documentId);
    }
```

I'm not at all expecting you to fix this! I was just wondering if maybe there is a reason why these have no types?

mmachatschek:
Hi @bidoubiwa

The thing is that if we would type these as string, we would break implementations that pass in documents with integer ids.

I looked at most places and found that either docblocks (which are more applicable here) are missing, incomplete or missing. I think I can send in another PR trying to cover all docblocks and types. Maybe I find more, where it makes sense to add specific types.

Cheers

bidoubiwa:
Thanks for the explaination :) Very cool if you do that but in either case thanks!

curquiza:
thanks @bidoubiwa and @mmachatschek!
bors merge

PR #170, opened 2021-05-15
Make composer lint command work

bidoubiwa:
Running `composer lint` locally would not work. In the CI it was not used as well, instead it was using the working command : 

```
vendor/bin/php-cs-fixer fix -v --config=.php-cs-fixer.dist.php --using-cache=no --dry-run --allow-risky=yes
```


bidoubiwa:
Closed for newer more efficient PR

PR #172, opened 2021-05-21
Fix composer lint command

bidoubiwa:
The lint composer command pointed to a non existing config file resulting in the following error: 
`  Cannot read config file "./.php_cs.dist".    `

Path has been updated. 
`composer lint` and `composer lint:fix` are now functional

I changed the scripts used in github actions with the ones from `composer` file.

I'm wondering why it wasn't used before. Am I relying on composer when I shouln't be?


curquiza (approved):
thanks @bidoubiwa! 😄

bidoubiwa:
bors merge

bidoubiwa:
bors merge

PR #174, opened 2021-05-26
Fix ApiException constructor crashing when HTTP body is empty

### LINKED ISSUE #171: Frequent crashes with "Wrong parameters for MeiliSearch\Exceptions\ApiException"
As ironic as it is, it seems ApiException creation fails due to invalid parameters and crashes.

Here's an image of the stack trace from Sentry:
image

Seems like Client.php uses invalid order for the params in parseResponse error handler.

---- Update:
Nevermind, it seems like the order of parameters is correct for the ApiException constructor. I'll try to investigate further.

On a sidenote: we're using PHP 7.4.

Tarpsvo:
Fixes #171

This PR includes:
- Reworked ApiException constructor, the first argument is now of `ResponseInterface` type
- Fixed all uses of `ApiException`
- Changed `ApiException` tests to use the new constructor
- Added new test that validates that ApiException constructor does not crash with an empty response

### DIFF
--- a/src/Exceptions/ApiException.php
+++ b/src/Exceptions/ApiException.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace MeiliSearch\Exceptions;
 
 use Exception;
+use Psr\Http\Message\ResponseInterface;
 
 class ApiException extends Exception
 {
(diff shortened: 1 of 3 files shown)


bidoubiwa (approved):
LGTM 🔥 Thanks a lot

curquiza (approved):
perfect! Thanks a lot!

bors merge

PR #176, opened 2021-05-31
Fix logging for ApiException

curquiza:
oops 😇

### DIFF
--- a/src/Exceptions/ApiException.php
+++ b/src/Exceptions/ApiException.php
@@ -30,7 +30,7 @@ class ApiException extends Exception
 
     public function __toString()
     {
-        $base = 'MeiliSearch HTTPRequestException: Http Status: '.$this->httpStatus;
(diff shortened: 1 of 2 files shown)


bidoubiwa (dismissed):
bors merge

bidoubiwa:
bors -r

curquiza:
My bad! I fixed the tests 😇

bidoubiwa:
Sorry if i'm mistaken, but since the error name is different is this not a breaking change for users that tries to catch that specific error?

curquiza:
It's only the displayed logs when the error is raised, the name (in the code) is still the same -> `MeiliSearchApiException`, so the code of the users will not be breaking.
From my POV changing this kind of logs, I mean used during development, is not breaking for the usage. But if you would rather consider this kind of change as breaking, be free to use the `breaking change` label 🙂

bidoubiwa (approved):
I see! It's oke for me.

bidoubiwa:
bors merge

PR #181, opened 2021-06-29
Exception when improper utf8 chars

### LINKED ISSUE #180: Http requests use json_encode which can fail and cause a TypeError.
**Description**
If you send content to be updated or inserted with the Client, it uses `json_encode` before sending the request. If the content is not UTF-8 encoded... in my case I believe it was from using a non-english language, then parsing the content to json can fail. Take a look at the post method for example:

```php
    public function post($path, $body = null, $query = [])
    {
        $request = $this->requestFactory->createRequest(
            'POST',
            $this->baseUrl.$path.$this->buildQueryString($query)
        )->withBody($this->streamFactory->createStream(json_encode($body)));

        return $this->execute($request);
    }
```

The POST, PUT, and PATCH methods all use `createStream()` with `json_encode`.

When `json_encode` fails it just returns `false`... but the interface method `createStream(string $content = '')` requires content to be a string, so when it encounters a boolean from the failed json encoding it crashes with a `TypeError`.

The functions need to catch the error and maybe use a custom error message with a warning, or use a private function to check for encoding errors before uploading: 

```php
        try {
            $stream = $this->streamFactory->createStream(json_encode($body));
        } catch (\TypeError $e) {
            throw new \TypeError('Failed to parse content... make sure it's UTF-8 encoded.');
        }
```

If it fails you can also call `json_last_error()` and you'll get an INT which explains reason [why the json parser failed](https://www.php.net/manual/en/function.json-last-error.php#119958). In my code I got a `5: JSON_ERROR_UTF8` but as you can see there are a few reasons why the parser can fail.

**Current behavior**
This is the current error I receive:

```php
  Nyholm\Psr7\Factory\Psr17Factory::createStream(): Argument #1 ($content) must be of type string, bool given, called in /Users/daniel/Code/www.llresearch/vendor/meilisearch/meilisearch-php/src/Http/Client.php on line 111

  at vendor/nyholm/psr7/src/Factory/Psr17Factory.php:33
     29▕
     30▕         return new Response($code, [], null, '1.1', $reasonPhrase);
     31▕     }
     32▕
  ➜  33▕     public function createStream(string $content = ''): StreamInterface
     34▕     {
     35▕         return Stream::create($content);
     36▕     }
     37▕
```
     
**Temporary solution**
I was able to discover and overcome the error by forcing UTF-8 encoding, however it took a whole afternoon to debug so a warning would be useful. One reason why it took a long time to debug is that it threw a `TypeError` instead of an `Exception`.

```php
    // $document["content"] = $content;
    $document["content"] = utf8_encode($content);
```

codedge:
Fix #180 

This PR adds some more tests for properly checking the import of documtens with special characters like they occur in German, French or Russian.

Further it adds a check whether the `json_encode`, that is called to encode the payload sent to MS, fails and throw an exception.

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -9,6 +9,7 @@ use Http\Discovery\Psr18ClientDiscovery;
 use MeiliSearch\Contracts\Http;
 use MeiliSearch\Exceptions\ApiException;
 use MeiliSearch\Exceptions\CommunicationException;
+use MeiliSearch\Exceptions\FailedJsonEncodingException;
 use Psr\Http\Client\ClientExceptionInterface;
 use Psr\Http\Client\ClientInterface;
 use Psr\Http\Client\NetworkExceptionInterface;
@@ -84,21 +85,27 @@ class Client implements Http
     }
 
     /**
-     * @param string $path
-     * @param null   $body
-     * @param array  $query
+     * @param null|mixed $body
      *
      * @return mixed
      *
-     * @throws ClientExceptionInterface
      * @throws ApiException
+     * @throws ClientExceptionInterface
+     * @throws CommunicationException
+     * @throws FailedJsonEncodingException
      */
-    public function post($path, $body = null, $query = [])
+    public function post(string $path, $body = null, array $query = [])
     {
+        $content = json_encode($body);
+
+        if($content === false) {
+            throw new FailedJsonEncodingException('Encoding payload to json failed. ' . json_last_error_msg());
+        }
+
         $request = $this->requestFactory->createRequest(
             'POST',
             $this->baseUrl.$path.$this->buildQueryString($query)
-        )->withBody($this->streamFactory->createStream(json_encode($body)));
+        )->withBody($this->streamFactory->createStream($content));
 
         return $this->execute($request);
     }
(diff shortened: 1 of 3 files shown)


alallema (approved):
Thank you so much for your work @codedge on this PR.
Really nice with the tests.

alallema:
bors merge

src/Http/Client.php
```diff
@@ -84,21 +85,27 @@ public function get($path, $query = [])
...
-     * @throws ClientExceptionInterface
      * @throws ApiException
+     * @throws ClientExceptionInterface
+     * @throws CommunicationException
+     * @throws FailedJsonEncodingException
      */
-    public function post($path, $body = null, $query = [])
+    public function post(string $path, $body = null, array $query = [])
     {
+        $content = json_encode($body);
```
norkunas, line 99: Better late than never :) I'd suggest instead to leverage the `JSON_THROW_ON_ERROR`

codedge: Good point. I guess a PR is more than welcome 😄

alallema: @norkunas thanks for the catch!
I agreed with @codedge if you are comfortable with the idea of making a pr it would indeed be very welcome.

norkunas: in progress

norkunas: #185 :)

PR #217, opened 2021-10-04
Throw ApiException for invalid JSON responses in HTTP Client

### LINKED ISSUE #211: The default HTTP client can't handle non-JSON response bodies
**Description**

I just went through a bit of painful debugging because I couldn't figure out why my indexing POST requests were failing in production. This is the error message I got:

> MeiliSearch\Exceptions\FailedJsonDecodingException
> Decoding json payload failed: "Syntax error".
> Caused by: JsonException 

The error came from `MeiliSearch\Http\Client::parseResponse()` which uses `MeiliSearch\Http\Client::jsonDecode()`.

Now the actual problem was that in production, the MeiliSearch instance is sitting behind an nginx reverse proxy and the POST request was hitting the `max_client_body_size`, so nginx returned a 413 error with the default HTML error page.

This was difficult to debug since it only happened in production (I use the docker image for testing, so no reverse proxy in the test environment).

**Expected behavior**

The `Client::parseResponse` method should be able to handle non-JSON responses more gracefully so that a failing request in production produces a helpful error message. The client could either check the `Content-Type` header of the response and only parse it as JSON if appropriate. Or catch the exception thrown by `jsonDecode` and throw a more helpful exception with the HTTP response code.

**Current behavior**

If the MeiliSearch server returns a non-JSON response, you get a `FailedJsonDecodingException` with no indication regarding the actual response code and content.

**Environment**

- OS: Ubuntu 20.04
- MeiliSearch version: v.0.22.0
- meilisearch-php version: v0.19.1

srichter:
This will handle a `FailedJsonDecodingException` and convert it to an `ApiException` to provide more context (the HTTP status code).

If you think it will make more sense to check the `Content-Type` header I can refactor to use that.

Resolves #211

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -193,13 +193,18 @@ class Client implements Http
             return null;
         }
 
-        if ($response->getStatusCode() >= 300) {
-            $body = $this->jsonDecode($response->getBody()->getContents()) ?? $response->getReasonPhrase();
+        try {
+            $body = $this->jsonDecode($response->getBody()->getContents());
+        } catch (FailedJsonDecodingException $e) {
+            throw new ApiException($response, ['message' => $e->getMessage()], $e);
+        }
 
+        if ($response->getStatusCode() >= 300) {
+            $body = $body ?? $response->getReasonPhrase();
             throw new ApiException($response, $body);
         }
 
-        return $this->jsonDecode($response->getBody()->getContents());
+        return $body;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


alallema (commented):
Thanks for this PR @srichter! And sorry for the delay.
It is already more clear thanks but the `ApiException` is for return the error from the `Meilisearch` API, of this [type](https://docs.meilisearch.com/errors/#meilisearch-errors). I think it's best to keep `FailedJsonDecodingException` and modify it or create a new one.
Also if you can add the `content-type` it would be amazing.
Thanks a lot

srichter:
Oh, that makes sense it has a specific meaning. I've created a new PR #219  with the alternate implementation.

PR #219, opened 2021-10-05
Check response Content-Type header in HTTP client

### LINKED ISSUE #211: The default HTTP client can't handle non-JSON response bodies

srichter:
Resolves #211 

This implements a new exception `InvalidResponseBodyException` which is thrown when parsing an HTTP response which does not have the Content-Type header `application/json`.

Based on my testing and research, the Meilisearch server will always send `application/json` as the Content-Type except on responses where the status code is `204`. The check is naive and assumes that the Meilisearch server will stick to this convention. If we want the check to be more robust (i.e. case-insensitive, able to handle charset being specified) I can adjust it as necessary.

### DIFF
--- a/src/Exceptions/InvalidResponseBodyException.php
+++ b/src/Exceptions/InvalidResponseBodyException.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MeiliSearch\Exceptions;
+
+use Exception;
+use Psr\Http\Message\ResponseInterface;
+
+class InvalidResponseBodyException extends Exception
+{
+    public $httpStatus = 0;
+    public $httpBody = null;
+    public $message = null;
+
+    public function __construct(ResponseInterface $response, $httpBody, $previous = null)
+    {
+        $this->httpStatus = $response->getStatusCode();
+        $this->httpBody = $httpBody;
+        $this->message = $this->getMessageFromHttpBody() ?? $response->getReasonPhrase();
+
+        parent::__construct($this->message, $this->httpStatus, $previous);
+    }
+
+    public function __toString()
+    {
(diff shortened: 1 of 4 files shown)


alallema (approved):
Everything looks good to me. Thanks for this PR @srichter 🔥 

Thank you for contributing to MeiliSearch. If you are participating in [Hacktoberfest](https://blog.meilisearch.com/contribute-hacktoberfest-2021/), and you would like to receive a small gift from MeiliSearch too, please complete [this form](https://meilisearch.typeform.com/to/rPexR5gq).

alallema:
bors merge

PR #226, opened 2021-10-06
Add deleteIndexIfExists method

### LINKED ISSUE #224: Add a method to delete the index only if it already exists
Trying to delete an index that does not exist on the MeiliSearch server will throw an error. To delete it without any error no matter what if it exists or not, we need a new method.

- [ ] Add a method `delete_index_if_exists`: it deletes the index but does not throw any error if the index does not exist.
- [ ] Add tests

Example: https://github.com/meilisearch/meilisearch-python/pull/268/files

⚠️ This issue is generated, and `delete_index_if_exists` might be named differently in this package. Keep the already existing way of naming in this package to stay idomatic with the language and the repository.

Related to https://github.com/meilisearch/integration-guides/issues/107

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue.

bofalke:
Closes #224

### DIFF
--- a/src/Delegates/HandlesIndex.php
+++ b/src/Delegates/HandlesIndex.php
@@ -35,6 +35,20 @@ trait HandlesIndex
         return $this->index($uid)->delete();
     }
 
+    public function deleteIndexIfExists(string $uid): bool
+    {
+        try {
+            $this->deleteIndex($uid);
+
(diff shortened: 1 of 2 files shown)


alallema:
bors merge

PR #237, opened 2021-10-13
Add createdAt and updatedAt to Indexes

### LINKED ISSUE #231: Add `updatedAt` and `createdAt` to `Index`
When manipulating indexes (get/create/update), the MeiliSearch server returns the following information related to the index:
- `uid`: the uid of the index
- `primaryKey`: the primary key of the index
- `updateAt`: when the index has been updated
- `createdAt`: when the index has been created

This SDK, when manipulating indexes (get/create/update) returns an instance of `Index` which already contains the `primaryKey` and `uid` members.

It means we can currently get the `primaryKey` and the `uid`, but not `updatedAt` and `createdAt`.

TODO:
- [ ] Add `updatedAt` and `createdAt` member in the `Index` class
- [ ] Update the variable in the code where `primaryKey` is already updated. Ex: when getting the index with `getIndex()`
- [ ] Add tests

⚠️ This issue is generated, it means `primaryKey`, `updatedAt`, and `createdAt` might be named differently in this package (ex: `primary_key` instead). Keep the already existing way of naming in this package to stay idiomatic with the language and the repository.
Also, if you are a maintainer, feel free to add any clarification and instruction about this issue related to the specific languages of the repo.

Related to https://github.com/meilisearch/integration-guides/issues/121

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue in the repo.

srichter:
Resolves #231.

This adds `$createdAt` and `$updatedAt` as properties to `Endpoints\Indexes`. New methods are added for retrieving the dates as `DateTime` instances or as raw strings (`getCreatedAt()`, `getCreatedAtString()`).

The `parseDate` static method is used because the Meilisearch server returns dates with 9 digits of fractional seconds, but `DateTime` can only parse dates containing up to 8 digits. 

I've also included a commit which refactors `Endpoints\Indexes` to use two new methods (`newInstance()` and `fill()`) to reduce code duplication when creating/updating instances from API responses. If you think this change should be a separate PR I can split it off.

### DIFF
--- a/src/Delegates/HandlesIndex.php
+++ b/src/Delegates/HandlesIndex.php
@@ -25,6 +25,11 @@ trait HandlesIndex
         return $this->index->allRaw();
     }
 
+    public function getRawIndex(string $uid): array
+    {
+        return $this->index($uid)->fetchRawInfo();
+    }
+
     public function index(string $uid): Indexes
     {
         return new Indexes($this->http, $uid);
--- a/tests/Endpoints/ClientTest.php
+++ b/tests/Endpoints/ClientTest.php
@@ -27,6 +27,15 @@ final class ClientTest extends TestCase
         $this->assertEmpty($response);
     }
 
+    public function testExceptionIsThrownOnGetRawIndexWhenIndexDoesNotExist(): void
+    {
+        $this->expectException(ApiException::class);
+        $this->expectExceptionMessage('Index index not found');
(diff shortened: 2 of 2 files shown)


srichter:
Pushed one more commit that removes an erroneous docblock. Figured I'd catch it before the merge rather than open another PR.

srichter:
It's good to go

PR #268, opened 2022-01-03
Redesign update API to task API

alallema:
### Breaking Changes
All the actions on indexes are now asynchronous:
- `createIndex()`, `updateIndex()`, `deleteIndex()` return a `task` response instead of an `Index`.
- `index.create` and `index.delete` from index return a `task`.
- `waitForPendingUpdate()`
    - is renamed into `waitForTask`
    - the current `index.waitForTask()` method call `/tasks/:uid`
- `index.getUpdateStatus` is renamed `index.getTask`
- `index.getAllUpdateStatus` is renamed `index.getAllTasks`

**Notes:** The only two methods that now return an `Index` are `client.index()` and `client.getIndex()`

### Enhancements
- new method `client.waitForTask()` call `/tasks/:uid`
- new method `client.waitForAllTask()`
- new method `client.getAllTasks` that calls `/tasks`
- new method `client.getTask` that calls `/tasks/:uid`


src/Endpoints/Indexes.php
```diff
@@ -152,49 +157,26 @@ public function fetchInfo(): self
...
-        while ($timeoutInMs > $timeout_temp) {
-            $res = $this->getUpdateStatus($updateId);
-            if ('enqueued' != $res['status'] && 'processing' != $res['status']) {
-                return $res;
-            }
-            $timeout_temp += $intervalInMs;
-            usleep(1000 * $intervalInMs);
-        }
-        throw new TimeOutException();
+        return $this->tasks->allIndexTask(self::PATH.'/'.$this->uid);
```
bidoubiwa, line 179: to be as consistent as possible with the other routes I would suggest the following naming: 

```
  getClientTask
  getClientTasks
  getIndexTask
  getIndexTasks
```

src/Endpoints/Indexes.php
```diff
@@ -152,49 +157,26 @@ public function fetchInfo(): self
...
 
-    public function getUpdateStatus($updateId): array
+    public function getTask($uid): array
     {
-        return $this->http->get(self::PATH.'/'.$this->uid.'/updates/'.$updateId);
+        return $this->tasks->getIndexTask(self::PATH.'/'.$this->uid, $uid);
     }
 
-    public function getAllUpdateStatus(): array
+    public function getAllTasks(): array
```
bidoubiwa, line 177: I think we decided to go for `getTasks`

src/Endpoints/Tasks.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MeiliSearch\Endpoints;
+
+use MeiliSearch\Contracts\Endpoint;
+use MeiliSearch\Exceptions\TimeOutException;
+
+class Tasks extends Endpoint
```
bidoubiwa, line 10: Since these do not match the naming used above, I would create to file in `endpoints`
`ClientTasks.php` and `IndexTasks.php`

src/Endpoints/Delegates/HandlesTasks.php
```diff
@@ -0,0 +1,33 @@
...
+declare(strict_types=1);
+
+namespace MeiliSearch\Endpoints\Delegates;
+
+use MeiliSearch\Endpoints\Tasks;
+
+/**
+ * @property Tasks tasks
+ */
+trait HandlesTasks
```
bidoubiwa, line 12: These are only used for the client?

src/Endpoints/Tasks.php
bidoubiwa: another option would be to be able to do 
```php
$this->tasks->index->getTasks(self::PATH.'/'.$this->uid);

# OR
$this->tasks->client->getTasks(self::PATH.'/'.$this->uid);
```

src/Endpoints/Tasks.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * @param array $uids
+     * @param int   $timeoutInMs
+     * @param int   $intervalInMs
+     *
+     * @throws TimeOutException
+     */
+    public function waitAllTasks($uids, $timeoutInMs, $intervalInMs): array
```
bidoubiwa, line 62: all uses of `All` in the methods should be removed

src/Endpoints/Delegates/HandlesTasks.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        return $this->tasks->get($uid);
+    }
+
+    public function getAllTasks(): array
+    {
+        return $this->tasks->all();
+    }
+
+    public function waitForTask($uid, $timeoutInMs = 5000, $intervalInMs = 50): array
```
bidoubiwa, line 24: maybe the parameter should be called `taskUid` to be sure that it is not confused with `indexUid`

src/Endpoints/Indexes.php
```diff
@@ -152,49 +157,26 @@ public function fetchInfo(): self
...
     }
 
-    // Updates
+    // Tasks
 
-    public function getUpdateStatus($updateId): array
+    public function getTask($uid): array
     {
-        return $this->http->get(self::PATH.'/'.$this->uid.'/updates/'.$updateId);
+        return $this->tasks->getIndexTask(self::PATH.'/'.$this->uid, $uid);
```
bidoubiwa, line 174: An inspiration for this could be the `stats` route that is also used in both client and index.

bidoubiwa (commented):
This is how [I did it in javascript](https://github.com/meilisearch/meilisearch-js/pull/1108/files#diff-2845baa1ab78b5edc8e5e46d1980d39eac8367b81302b1715809ef9545e42373). with the idea to keep them close together in case of changes. Nonetheless I do not guarantee this is the way to go. 
I made some suggestions in your code but the final decision is not clear for me neither:

src/Endpoints/Indexes.php
alallema: Yep but there is the function `getAllIndexes` and `deleteAllIndexes` so I change it to be consistent but there is also `stats` and `getSettings`.

src/Endpoints/Delegates/HandlesTasks.php
alallema: Yes

bidoubiwa (approved):
LGTM 🔥

PR #281, opened 2022-01-13
Make settings and synonyms be classes so they can be serialized properly

### LINKED ISSUE #209: Implementation of a Json Serializer
**Description**
As a result of this [issue](https://github.com/meilisearch/meilisearch-php/issues/204), it appears that `json_decode` transforms
```json
"synonyms": {}
```
into
```php
["synonyms"]=>
  array(0) {
}
```
In the same way as:
```json
"stopWords": []
```
become:
```php
["stopWords"]=>
  array(0) {
}
```
The problem appears when you send back the synonyms array to the MeiliSearch API who only take:
```json
"synonyms": {},
```
or
```json
"synonyms": null,
```
It could be useful to implement our own specific JsonSerializer to check if every data send back to MeiliSearch API are well-formatted.

In the meanwhile a patch will be created [see](https://github.com/meilisearch/meilisearch-php/pull/208).

jonatanrdsantos:
This pull request introduces 2 new classes to the Project: Settings and Synonyms.
Both extend from Data, the Data class act as a container to support the other 2 so we can have array access.

# Pull Request

## What does this PR do?
Fixes #209

## PR checklist

Thank you so much for contributing to MeiliSearch!

### DIFF
--- a/src/Endpoints/Delegates/HandlesSettings.php
+++ b/src/Endpoints/Delegates/HandlesSettings.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace MeiliSearch\Endpoints\Delegates;
 
+use MeiliSearch\Contracts\Index\Synonyms;
+
 trait HandlesSettings
 {
     // Settings - Ranking rules
@@ -95,18 +97,13 @@ trait HandlesSettings
 
     public function getSynonyms(): array
     {
-        return $this->http->get(self::PATH.'/'.$this->uid.'/settings/synonyms');
+        return (new Synonyms($this->http->get(self::PATH.'/'.$this->uid.'/settings/synonyms')))
+            ->getIterator()->getArrayCopy();
     }
 
     public function updateSynonyms(array $synonyms): array
     {
-        // Patch related to https://github.com/meilisearch/meilisearch-php/issues/204.
-        // Should be removed when implementing https://github.com/meilisearch/meilisearch-php/issues/209
-        if (0 === \count($synonyms)) {
-            $synonyms = null;
-        }
-
-        return $this->http->post(self::PATH.'/'.$this->uid.'/settings/synonyms', $synonyms);
+        return $this->http->post(self::PATH.'/'.$this->uid.'/settings/synonyms', new Synonyms($synonyms));
     }
 
     public function resetSynonyms(): array
--- a/src/Endpoints/Indexes.php
+++ b/src/Endpoints/Indexes.php
@@ -8,6 +8,7 @@ use DateTime;
 use Exception;
 use MeiliSearch\Contracts\Endpoint;
 use MeiliSearch\Contracts\Http;
+use MeiliSearch\Contracts\Index\Settings;
 use MeiliSearch\Endpoints\Delegates\HandlesDocuments;
 use MeiliSearch\Endpoints\Delegates\HandlesSettings;
 use MeiliSearch\Endpoints\Delegates\HandlesTasks;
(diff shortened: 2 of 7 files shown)


alallema:
Hi @jonatanrdsantos,
Thank you so much for this PR and sorry for the delay. I will try to review your PR today!

src/Endpoints/Delegates/HandlesSettings.php
```diff
@@ -98,20 +99,14 @@ public function resetStopWords(): array
 
     // Settings - Synonyms
 
-    public function getSynonyms(): array
+    public function getSynonyms(): Synonyms
     {
-        return $this->http->get(self::PATH.'/'.$this->uid.'/settings/synonyms');
+        return new Synonyms($this->http->get(self::PATH.'/'.$this->uid.'/settings/synonyms'));
```
mmachatschek, line 104: I would prefer to get an array instead of an Synonyms object here. This change would break the code of users that expect an array.
We could return the array form like this:

```suggestion
        return (new Synonyms($this->http->get(self::PATH.'/'.$this->uid.'/settings/synonyms')))->getIterator()->getArrayCopy();
```

src/Endpoints/Indexes.php
```diff
@@ -219,19 +220,13 @@ public function stats(): array
 
     // Settings - Global
 
-    public function getSettings(): array
+    public function getSettings(): Settings
     {
-        return $this->http->get(self::PATH.'/'.$this->uid.'/settings');
+        return new Settings($this->http->get(self::PATH.'/'.$this->uid.'/settings'));
```
mmachatschek, line 225: I would prefer to get an array instead of an Synonyms object here. This change would break the code of users that expect an array.
We could return the array form like this:

```suggestion
        return (new Settings($this->http->get(self::PATH.'/'.$this->uid.'/settings')))->getIterator()->getArrayCopy();
```

alallema (commented):
Thanks for this PR @jonatanrdsantos 
I'm not going to add a review since @mmachatschek's review is already very relevant I think.
I have indeed the same question for the both interfaces `src/Contracts/Index/SettingsInterface.php` and `src/Contracts/Index/SynonymsInterface.php`

src/Endpoints/Delegates/HandlesSettings.php
jonatanrdsantos: Hi @mmachatschek regarding this, Synonyms being an ArrayAccess class will prevent the code to break, cause the client can loop and use it as an array right?

src/Endpoints/Indexes.php
jonatanrdsantos: Hi @mmachatschek regarding this, Settings being an ArrayAccess class will prevent the code to break, cause the client can loop and use it as an array right?

jonatanrdsantos:
@alallema and @mmachatschek Regarding the interface I prefer to program with interfaces always I can, but in this case, is very simple and there's no need for an interface.

Updated the code removing them.

On the classes Settings and Synonyms, they implement \ArrayAccess, \Countable, \IteratorAggregate, só te user can loop, integrate and work with as Array.

If is mandatory a scalar array type return I can update the code with no problem.

jonatanrdsantos:
@alallema @mmachatschek Can u please review, I applied the changes as required.

alallema (approved):
Hi @jonatanrdsantos,
Thank you so much for this PR and so sorry for the delay ❤️ 
LGTM! 🎉

alallema:
bors merge

PR #286, opened 2022-01-20
Update PHPstan to stable version 1

mmachatschek:
# Pull Request

## What does this PR do?
This updates to the current stable PHPStan version

## PR checklist
Please check if your PR fulfills the following requirements:
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

PR #287, opened 2022-01-20
Improve phpstan level

### LINKED ISSUE #182: Traits break contract for not reducing any complexity
The ways traits are used in this lib feel a bit strange. Take the 

* `HandlesDocuments` trait or
* `HandlesDumps` trait
* ... and many more...

Both are only used _once_. Further traits access properies 

* `HandlesDocuments` accesses `http`
* `HandlesDumps` accesses `dumps`

that are not declared inside the trait but in the class they are used in. Even with the `@property` notation added inside the traits, it is not clear to which property you are refering. Using the `@property` notation is meant for magic methods. 

Why using a trait for methods that are only used in _one_ place? Directly implementing would have the same effect and would provide a cleaner style.

I was coming across that when I wanted to increase PHPstans level to `2`. With the current implementation this seems not possible.

Let me know if the methods implemented in traits can be moved to the classes itself and/or how PHPstan can correctly determine the traits properties _that are not defined inside the trait itself_.

grafik

mmachatschek:
# Pull Request

## What does this PR do?
Fixes #182

I split up each level into a seperate commit to follow easier.

I didn't increase to level > 5 because we would need to add specific typed docblocks to each method that returns an array. It could be handy in the future, but I think there are better ways to solve that.

## PR checklist

This is in Draft mode because it depends on PR #286

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,5 @@
 parameters:
-    level: 1
+    level: 2
     paths:
         - src
         - tests
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,5 @@
 parameters:
-    level: 2
+    level: 3
     paths:
         - src
         - tests
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,5 @@
 parameters:
-    level: 4
+    level: 5
     paths:
         - src
         - tests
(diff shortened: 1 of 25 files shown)


alallema (approved):
Thanks, @mmachatschek for the fixe and all the cleaning 🎉 
LGTM!
And thanks for contributing to MeiliSearch ❤️

alallema:
bors merge

PR #298, opened 2022-02-21
Changes the Keys Class

alallema:
### Breaking Changes

The `class` `Keys` wasn't used as a proper class and did not return objects from the `Keys` but arrays. This has been changed. Setters and Getters have been added to the class.

- `client.getKey` return a Keys
- `client.createKey` return a Keys
- `client.updateKey` return a Keys

### DIFF
--- a/src/Endpoints/Keys.php
+++ b/src/Endpoints/Keys.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace MeiliSearch\Endpoints;
 
+use DateTime;
 use MeiliSearch\Contracts\Endpoint;
 use MeiliSearch\Contracts\Http;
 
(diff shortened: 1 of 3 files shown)


src/Endpoints/Keys.php
```diff
@@ -5,29 +5,133 @@
...
+            $attributes['expiresAt'],
+            $attributes['createdAt'],
+            $attributes['updatedAt'],
+        );
+    }
+
+    /**
+     * @return $this
+     */
+    protected function fill(array $attributes): self
```
brunoocasali, line 52: I'm not sure how it should be done, but it seems to me like a duplication, what about try to reuse this `fill` with the `newInstance`?

brunoocasali (commented):
Well done! 🥇

src/Endpoints/Keys.php
alallema: Yes it was I remove it

bidoubiwa (approved):
LGTM 🔥

alallema:
bors merge

brunoocasali:
@alallema is there any particular reason to merge this into the `bump` instead of main?

PR #299, opened 2022-02-27
Modify Client test to check ApiException code rather than ApiException message

### LINKED ISSUE #252: Test the `code` of the `ApiException` exception instead of the `message`
In tests, when an `ApiException` exception is tested, often the message of the exception is also checked.

```php
$this->expectException(ApiException::class);
$this->expectExceptionMessage('internal error');
```

This is something we don't want to do. We only want to check the `code` of the error in the tests, not the `message` itself that can often break on the MeiliSearch side.

FYI, MeilISearch returns 4 fields when raising an error: `message`, `code`, `type`, and `link`. Only the `code` should be tested.

⚠️ Should be done once MeiliSearch v0.24.0 is released

TeddyBear06:
Check ApiException error code rather than error message to avoid tests break due to Meilisearch API evolving.

# Pull Request

## What does this PR do?
Fixes #252 

## PR checklist

### DIFF
--- a/tests/Http/ClientTest.php
+++ b/tests/Http/ClientTest.php
@@ -72,7 +72,7 @@ class ClientTest extends TestCase
         $client = new Client('https://localhost', null, $httpClient);
 
         $this->expectException(ApiException::class);
-        $this->expectExceptionMessage('internal error');
+        $this->expectExceptionCode(300);
 
         $client->post('/', '');
     }
@@ -119,7 +119,7 @@ class ClientTest extends TestCase
         $client = new Client('https://localhost', null, $httpClient);
 
         $this->expectException(ApiException::class);
-        $this->expectExceptionMessage('internal error');
+        $this->expectExceptionCode(300);
 
         $client->put('/', '');
     }
@@ -166,7 +166,7 @@ class ClientTest extends TestCase
         $client = new Client('https://localhost', null, $httpClient);
 
         $this->expectException(ApiException::class);
-        $this->expectExceptionMessage('internal error');
+        $this->expectExceptionCode(300);
 
         $client->post('/', '');
     }
--- a/tests/Http/ClientTest.php
+++ b/tests/Http/ClientTest.php
@@ -67,14 +67,14 @@ class ClientTest extends TestCase
 
     public function testPostThrowsApiException(): void
     {
-        $httpClient = $this->createHttpClientMock(300, '{"message":"internal error"}');
-
-        $client = new Client('https://localhost', null, $httpClient);
-
-        $this->expectException(ApiException::class);
-        $this->expectExceptionCode(300);
-
-        $client->post('/', '');
+        try {
+            $httpClient = $this->createHttpClientMock(300, '{"message":"internal error","code":"internal"}');
+            $client = new Client('https://localhost', null, $httpClient);
+            $client->post('/', '');
+            $this->fail('ApiException not raised.');
+        } catch (ApiException $e) {
+            $this->assertEquals('internal', $e->errorCode);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


alallema (commented):
Thanks for this PR @TeddyBear06!
There are multiple things to modify but maybe the issue was not detailed enough. First, the error codes mentioned in the issue can be found [here](https://docs.meilisearch.com/reference/api/error_codes.html#error-list). Second all these tests are mock so the mock function should be modified to test the `error_code`:
```
    private function createHttpClientMock(int $status = 200, string $content = '{', string $contentType = 'application/json')
```
And all the error messages should be removed.

TeddyBear06:
Hi @alallema,

I totally missed [that point](https://github.com/meilisearch/meilisearch/blob/main/meilisearch-error/src/lib.rs#L18), "error_code" here is a string (it makes me think of a 3 digits int) :sweat_smile:

Ok, it makes more sense now.

But if I'm not mistaken, errorCode here is not "supported" by `ResponseInterface` nor `MessageInterface`, I mean we cannot use a `getErrorCode` so modifying [createHttpClientMock](https://github.com/meilisearch/meilisearch-php/blob/main/tests/Http/ClientTest.php#L256) may be a little bit tricky/hacky.

On the other side, we can modify tests like so:

```php
public function testPutThrowsApiException(): void
{
    try {
        $httpClient = $this->createHttpClientMock(300, '{"message":"internal error","code":"internal"}');
        $client = new Client('https://localhost', null, $httpClient);
        $client->put('/', '');
        $this->fail('ApiException not raised.');
    } catch (ApiException $e) {
        $this->assertEquals('internal', $e->errorCode);
    }
}
```

What do you think about it? If you have any clue about `createHttpClientMock` modification, let me know.

alallema:
> I totally missed [that point](https://github.com/meilisearch/meilisearch/blob/main/meilisearch-error/src/lib.rs#L18), "error_code" here is a string (it makes me think of a 3 digits int) 😅

For me too! The error_code seems to be a number indeed.

> On the other side, we can modify tests like so:
> ```
> public function testPutThrowsApiException(): void
> {
>     try {
>         $httpClient = $this->createHttpClientMock(300, '{"message":"internal error","code":"internal"}');
>         $client = new Client('https://localhost', null, $httpClient);
>         $client->put('/', '');
>         $this->fail('ApiException not raised.');
>     } catch (ApiException $e) {
>         $this->assertEquals('internal', $e->errorCode);
>     }
> }
> ```
> What do you think about it? If you have any clue about createHttpClientMock modification, let me know.

This seems great to me! Thanks again

TeddyBear06:
I had a look at failed tests and it seems to be related to [older untouched tests (mainly `Connection reset by peer`)](https://github.com/meilisearch/meilisearch-php/runs/5389583840?check_suite_focus=true#step:7:22) :sweat_smile:

BTW, thoose errors are interesting as the expected exception is not the one that was thrown (`Failed asserting that exception of type "GuzzleHttp\Exception\RequestException" matches expected exception "MeiliSearch\Exceptions\ApiException"`):

- https://github.com/meilisearch/meilisearch-php/runs/5389583840?check_suite_focus=true#step:7:501
- https://github.com/meilisearch/meilisearch-php/runs/5389583840?check_suite_focus=true#step:7:527
- https://github.com/meilisearch/meilisearch-php/runs/5389583840?check_suite_focus=true#step:7:554
- 
Maybe you could relauch tests suite?

alallema:
Hi @TeddyBear06,
Sorry for the delay, we are currently working on the next version which is the reason why there are problems with the tests.

alallema:
bors merge

PR #305, opened 2022-03-16
Fix authorization header when no api key is provide

### LINKED ISSUE #304: The provided API key is invalid.
Hello everyone,

after update this package to the newest version all my tests are failing with the following error:
The provided API key is invalid.

Am I missing something?
The MEILISEARCH_KEY is set to null on my config file.

Thank you in advance

alallema:
If no API key is provided for `Meilisearch` the Client added the header anyway.
fixes #304.

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -47,9 +47,11 @@ class Client implements Http
         $this->requestFactory = $reqFactory ?? Psr17FactoryDiscovery::findRequestFactory();
         $this->streamFactory = Psr17FactoryDiscovery::findStreamFactory();
         $this->headers = array_filter([
-            'Authorization' => sprintf('Bearer %s', $this->apiKey),
             'User-Agent' => MeiliSearch::qualifiedVersion(),
         ]);
+        if ($this->apiKey) {
+            $this->headers['Authorization'] = sprintf('Bearer %s', $this->apiKey);
+        }
         $this->json = new Json();
     }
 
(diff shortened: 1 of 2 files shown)


src/Http/Client.php
```diff
@@ -47,9 +47,11 @@ public function __construct(
         $this->requestFactory = $reqFactory ?? Psr17FactoryDiscovery::findRequestFactory();
         $this->streamFactory = Psr17FactoryDiscovery::findStreamFactory();
         $this->headers = array_filter([
-            'Authorization' => sprintf('Bearer %s', $this->apiKey),
             'User-Agent' => MeiliSearch::qualifiedVersion(),
         ]);
+        if ($this->apiKey) {
```
bidoubiwa, line 52: This is throwing an error on `phpstan`

> Only booleans are allowed in an if condition, string|null given.

bidoubiwa (changes requested):

Tylerian (commented):
I would add an extra test case when no apiKey is provided

bidoubiwa (approved):
LGTM !🔥

alallema:
bors merge

alallema:
bors merge

brunoocasali:
@alallema this is a valid behavior, isn't it? Since if we have in the development mode, Meilisearch server does allow us to start the server without providing a `masterKey`.

alallema:
@brunoocasali, yes this is the only case you can run your `Meilisearch` instance without master key.

PR #308, opened 2022-03-16
Update CONTRIBUTING.md

### LINKED ISSUE #307: Alter composer.json `allow-plugins` section
Hi,

Newcomers can experience some "frictions" starting contributing as discussed here : https://github.com/meilisearch/meilisearch-php/issues/252#issuecomment-1054051683

There is still an issue regarding [phpstan using Composer 2.2.x](https://github.com/phpstan/extension-installer/pull/41/files) as `d` means `discard` not `download` and if we discard it, we cannot use phpstan:

```bash
Do you trust "phpstan/extension-installer" to execute code and wish to enable it now? (writes "allow-plugins" to composer.json) [y,n,d,?] ?
y - add package to allow-plugins in composer.json and let it run immediately
n - add package (as disallowed) to allow-plugins in composer.json to suppress further prompts
d - discard this, do not change composer.json and do not allow the plugin to run
? - print help
```

We have several options:

1. Explicity ask devs to not use Composer 2.2.x (but the [Composer backward compatibility is until July 2022](https://getcomposer.org/doc/06-config.md#allow-plugins))
2. Ask contributors to take the responsability to allow plugins and revert changes on composer.json
3. Add a [conflict section](https://getcomposer.org/doc/04-schema.md#conflict) to the composer.json:

```json
"conflict": {
  "composer/composer": "~2.2"
}
```

4. Add phpstan to the `allow-plugins` list into the composer.json like so:

```json
{
    "config": {
        "allow-plugins": {
            "phpstan/extension-installer": true
        }
    }
}
```

But honestly, I think that:

- 1 is not a good one as we cannot use brakes on updates
- 2 seems not that bad from a security perspective as it's the user who decide wich plugin to trust
- 3 is a temporary - kind of - fix
- 4 imply security considerations out of the scope of Meilisearch PHP (I mean, we need to give an explicit trust to phpstan silently (with all the respect that I've for phpstan)) but I think that it's what folks at Composer wanted

I really need your opinion on this issue.

TeddyBear06:
# Pull Request

## What does this PR do?
Fixes #307 

## PR checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,8 +7,17 @@ First of all, thank you for contributing to Meilisearch! The goal of this docume
 - [Assumptions](#assumptions)
 - [How to Contribute](#how-to-contribute)
 - [Development Workflow](#development-workflow)
+  - [Requirements](#requirements)
+  - [Setup](#setup)
+  - [Tests and Linter](#tests-and-linter)
 - [Git Guidelines](#git-guidelines)
+  - [Git Branches](#git-branches)
+  - [Git Commits](#git-commits)
+  - [GitHub Pull Requests](#github-pull-requests)
 - [Release Process (for internal team only)](#release-process-for-internal-team-only)
+  - [Automation to Rebase and Merge the PRs](#automation-to-rebase-and-merge-the-prs)
+  - [Automated Changelogs](#automated-changelogs)
+  - [How to Publish the Release](#how-to-publish-the-release)
 
 <!-- /MarkdownTOC -->
 
@@ -30,10 +39,26 @@ First of all, thank you for contributing to Meilisearch! The goal of this docume
 
 ## Development Workflow
 
+### Requirements
+
+- [Composer 2.x](https://getcomposer.org/download/)
+
 ### Setup
 
+Clone the project then:
+
 ```bash
+cd meilisearch-php
 composer install
+```
+
+If you don't already have a Meilisearch server instance running, execute the following commands:
+
+NB: Will create a bunch of files and folders, better to use a dedicated folder outside of this project.
+
+```bash
+curl -L https://install.meilisearch.com | sh # download Meilisearch
+./meilisearch --master-key=masterKey --no-analytics # run Meilisearch
 ```
 
 ### Tests and Linter
@@ -42,8 +67,6 @@ Each PR should pass the tests and the linter to be accepted.
 
 ```bash
 # Tests
-curl -L https://install.meilisearch.com | sh # download Meilisearch
-./meilisearch --master-key=masterKey --no-analytics # run Meilisearch
 composer test
 # Linter (with auto-fix)
 composer lint:fix


CONTRIBUTING.md
```diff
@@ -30,20 +39,34 @@ First of all, thank you for contributing to Meilisearch! The goal of this docume
...
+Clone the project then:
+
 ```bash
+cd meilisearch-php
 composer install
 ```
 
+If you don't already have a Meilisearch server instance running, execute the following commands:
+
+NB: Will create a bunch of files and folders, better to use a dedicated folder outside of this project.
```
alallema, line 57: This will not create a bunch of files, only the directory `data.ms` (database files) whose path you can easily configure with an environment variable or a flag

CONTRIBUTING.md
```diff
@@ -7,8 +7,17 @@ First of all, thank you for contributing to Meilisearch! The goal of this docume
 - [Assumptions](#assumptions)
 - [How to Contribute](#how-to-contribute)
 - [Development Workflow](#development-workflow)
+  - [Requirements](#requirements)
+  - [Setup](#setup)
+  - [Tests and Linter](#tests-and-linter)
```
alallema, line 12: These three sections are really small I don't think a sub-menu is necessary

CONTRIBUTING.md
```diff
@@ -7,8 +7,17 @@ First of all, thank you for contributing to Meilisearch! The goal of this docume
...
 - [Git Guidelines](#git-guidelines)
+  - [Git Branches](#git-branches)
+  - [Git Commits](#git-commits)
+  - [GitHub Pull Requests](#github-pull-requests)
```
alallema, line 16: Same as above

CONTRIBUTING.md
```diff
@@ -7,8 +7,17 @@ First of all, thank you for contributing to Meilisearch! The goal of this docume
...
 - [Release Process (for internal team only)](#release-process-for-internal-team-only)
+  - [Automation to Rebase and Merge the PRs](#automation-to-rebase-and-merge-the-prs)
+  - [Automated Changelogs](#automated-changelogs)
+  - [How to Publish the Release](#how-to-publish-the-release)
```
alallema, line 20: Same as above

CONTRIBUTING.md
```diff
@@ -30,20 +39,34 @@ First of all, thank you for contributing to Meilisearch! The goal of this docume
...
 
+If you don't already have a Meilisearch server instance running, execute the following commands:
+
+NB: Will create a bunch of files and folders, better to use a dedicated folder outside of this project.
+
+```bash
+curl -L https://install.meilisearch.com | sh # download Meilisearch
+./meilisearch --master-key=masterKey --no-analytics # run Meilisearch
+```
+
```
alallema, line 63: This command belong to the Test and Linter section because you don't need it to setup the project you will need it if you want launch the tests

alallema (commented):
Thank you so much for this PR @TeddyBear06 😍 
I left some comments also it could be a good idea to add the suggestion you made on this issue #307

CONTRIBUTING.md
TeddyBear06: You're right, just did read https://docs.meilisearch.com/learn/configuration/instance_options.html carefully.

I'll remove this part.

CONTRIBUTING.md
TeddyBear06: Ok, agreed.

TeddyBear06:
Putting it all together, I think that https://github.com/meilisearch/meilisearch-php/issues/307 solves the main onboarding "friction" and this PR does not add anything more than the existing version.

PR #315, opened 2022-04-21
Fix tests due to cropping and facet distribution

alallema:
As per the changes made in the cropping behavior (now per words see [this spec](https://github.com/meilisearch/specifications/blob/f5c6a8e183e22cc3b80d9a0251c411250f31674f/text/0118-search-api.md#3111-attributestocrop) tests needed to be fixed.


bidoubiwa (approved):
LGTM 🔥

PR #317, opened 2022-04-21
Ensure nested field support

alallema:
Add tests on nested fields to ensure it is still maintained in Meilisearch.


bidoubiwa (approved):
LGTM 🔥

PR #332, opened 2022-06-14
Apply search changes

curquiza:
- Rename `nbHits` response parameter to `estimatedTotalHits`.
- Delete `exhaustiveNbHits` response parameter.
- Delete `exhaustiveFacetsCount` response parameter.
- `matches` request parameter is renamed `showMatchesPosition`.
- `_matchesInfo` response parameter is renamed `_matchesPosition`.
- `facetsDistribution` request parameter is renamed `facets`.
- `facetsDistribution` response parameter is renamed `facetDistribution`.
- Rename `transformFacetsDistribution` into `transformFacetDistribution`
- Add a patch to "copy the value" from `estimatedTotalHits` to `nbHits`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ $index->search(
   ],
   "offset": 0,
   "limit": 20,
-  "nbHits": 1,
+  "estimatedTotalHits": 1,
   "processingTimeMs": 0,
   "query": "wonder"
 }
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -272,7 +272,7 @@ search_parameter_guide_highlight_tag_1: |-
     'highlightPostTag' => '</span>'
   ]);
 search_parameter_guide_matches_1: |-
-  $client->index('movies')->search('winter feast', ['attributesToHighlight' => ['overview'], 'matches' => true]);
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -207,7 +207,7 @@ $index->search(
   ],
   "offset": 0,
   "limit": 20,
-  "nbHits": 1,
```
brunoocasali, line 210: Just to remind myself, but we need to deal with the `nbHits` in the meilisearch-php to prevent the breaking in the laravel/scout!

brunoocasali (changes requested):
@curquiza no need to change anything, I would like to do this scout workaround :)

brunoocasali:
Add a patch fix to the result from `rawSearch`, to prevent laravel/scout breaking:

In scout, they call [`rawSearch`](https://github.com/laravel/scout/blob/9d37b622bc863d4f76539fce6c90b849b30a47ea/src/Engines/MeiliSearchEngine.php#L154):  and they refer as well to [`getTotalCount`](https://github.com/laravel/scout/blob/9d37b622bc863d4f76539fce6c90b849b30a47ea/src/Engines/MeiliSearchEngine.php#L297). 

So to prevent a breaking change there, I added this line: `$result['nbHits'] = $result['estimativeNbHits'];` before the result.

PR #335, opened 2022-06-29
Add missing docker command + supress warning

brunoocasali:
During the v0.28 implementation, I noticed that I had an annoying warning in the tests, and during the attempts of fixing this warning, I had to move to PHP 7.4 to test, doing that I noticed the Docker config didn't have the `zip` and `git` libraries that are required to composer work.

I probably miss that because the PHP repo saves the vendor stuff locally not in a volume as I was used to, this PR fixes both things!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,8 @@
 FROM php:8.1
 
+RUN apt-get update && apt-get install -y \
+    git \
+    unzip \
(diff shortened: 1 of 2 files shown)


alallema (approved):
LGTM! 🎉 
Like the detailed explanation in the description PR

brunoocasali:
bors merge

PR #337, opened 2022-06-29
Tasks refactor v1

brunoocasali:
This PR changes all the places that currently are handling the summarized task version from `uid` to `taskUid`. And replace the tasks route index namespace with the global one + add the ability to infer the index from it.

disclaimer: The work in the tasks is not done yet, this is the first step.

### DIFF
--- a/src/Endpoints/Indexes.php
+++ b/src/Endpoints/Indexes.php
@@ -151,12 +151,12 @@ class Indexes extends Endpoint
 
     public function getTask($uid): array
     {
-        return $this->http->get(self::PATH.'/'.$this->uid.'/tasks'.'/'.$uid);
+        return $this->http->get('/tasks/'.$uid, ['indexUid' => $this->uid]);
     }
 
     public function getTasks(): array
     {
-        return $this->http->get(self::PATH.'/'.$this->uid.'/tasks');
+        return $this->http->get('/tasks', ['indexUid' => $this->uid]);
     }
 
     // Search
--- a/tests/Endpoints/TasksTest.php
+++ b/tests/Endpoints/TasksTest.php
@@ -24,7 +24,7 @@ final class TasksTest extends TestCase
 
         $this->assertIsArray($response);
         $this->assertArrayHasKey('status', $response);
-        $this->assertSame($response['uid'], $promise['uid']);
+        $this->assertSame($response['uid'], $promise['taskUid']);
         $this->assertArrayHasKey('type', $response);
         $this->assertSame($response['type'], 'documentPartial');
         $this->assertArrayHasKey('indexUid', $response);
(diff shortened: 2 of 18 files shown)


tests/Endpoints/TasksTest.php
```diff
@@ -82,27 +83,30 @@ public function testGetOneTaskIndex(): void
...
+        $newFirstIndex = $response['results'][0]['uid'];
+
+        $this->assertEquals($firstIndex, $newFirstIndex);
     }
 
     public function testExceptionIfNoTaskIdWhenGetting(): void
     {
         $this->expectException(ApiException::class);
-        $this->index->getTask(9999999999);
+        $this->index->getTask(99999999);
```
brunoocasali, line 103: Reference: https://github.com/meilisearch/meilisearch/issues/2555

tests/Endpoints/TasksTest.php
```diff
@@ -82,27 +83,30 @@ public function testGetOneTaskIndex(): void
...
-        $preCount = \count($response['results']);
-        [$promise, $response] = $this->seedIndex();
+        $firstIndex = $response['results'][0]['uid'];
+
+        $newIndex = $this->createEmptyIndex('a_new_index');
+        $newIndex->updateDocuments(self::DOCUMENTS);
 
-        $this->assertIsArray($promise);
         $response = $this->index->getTasks();
-        $this->assertCount($preCount + 1, $response['results']);
```
brunoocasali, line 93: The test is basically the same, but now we cannot trust the size of the list anymore, so the idea is to just compare the most recent item if they changed, we can ensure that we are filtering by index. And in the other similar test the opposite is what we look for.

src/Endpoints/Indexes.php
```diff
@@ -151,12 +151,12 @@ public function delete(): array
 
     public function getTask($uid): array
     {
-        return $this->http->get(self::PATH.'/'.$this->uid.'/tasks'.'/'.$uid);
+        return $this->http->get('/tasks/'.$uid, ['indexUid' => $this->uid]);
```
alallema, line 154: On the general issue it specified the following point:
> Remove GET `/indexes/:indexUid/tasks/:taskUid`. Use GET `/tasks/:taskUid` instead.

So shouldn't it be something like
```suggestion
        return $this->http->get('/tasks'.'/'.$uid);
```

alallema (commented):
Well done ✨ !!!
Just one comment about the refactor of the `index.getTask(uid)`

src/Endpoints/Indexes.php
alallema: Even better you can use the one from the client maybe

brunoocasali: @alallema is there any particular reason to suggest:

```php
('/tasks'.'/'.$uid);
```
instead of:

```php
('/tasks/'.$uid);
```

brunoocasali:
@alallema I'll apply your suggestion in a next PR :)

PR #338, opened 2022-06-29
Keys changes v1

brunoocasali:
This PR:

- Adds two new attributes `name` and `uid` in tasks.
- Add $uid to `generateTenantToken` function.
- Fix code samples.
- ⚠️ Ignore silently the unused keys in the update `$options` array.

### DIFF
--- a/src/Endpoints/TenantToken.php
+++ b/src/Endpoints/TenantToken.php
@@ -36,7 +36,7 @@ class TenantToken extends Endpoint
      * - apiKey: The API key parent of the token. If you leave it empty the client API Key will be used.
      * - expiresAt: A DateTime when the key will expire. Note that if an expiresAt value is included it should be in UTC time.
      */
-    public function generateTenantToken($searchRules, ?array $options = []): string
+    public function generateTenantToken(string $uid, $searchRules, ?array $options = []): string
     {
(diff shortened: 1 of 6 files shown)


src/Endpoints/TenantToken.php
```diff
@@ -36,7 +36,7 @@ private function validateTenantTokenArguments($searchRules, ?array $options = []
      * - apiKey: The API key parent of the token. If you leave it empty the client API Key will be used.
      * - expiresAt: A DateTime when the key will expire. Note that if an expiresAt value is included it should be in UTC time.
      */
-    public function generateTenantToken($searchRules, ?array $options = []): string
+    public function generateTenantToken(string $uid, $searchRules, ?array $options = []): string
```
alallema, line 39: Like suggested in this [comment]
(https://github.com/meilisearch/meilisearch-python/pull/484#pullrequestreview-1017277409) it could be good to rename `uid` to `api_key_uid`

alallema (commented):
Well done @brunoocasali 🎉 
Just one comment about naming

src/Endpoints/TenantToken.php
brunoocasali: Thanks @alallema I'll apply this suggestion in a next PR, btw I will do it in the dart as well :)

PR #344, opened 2022-07-01
Documents changes v2

brunoocasali:
- Create `DocumentsQuery` to handle pagination in documents endpoint.
- Introduce `DocumentsResults` & make the `getDocuments` return a object
- Add `DocumentsQuery/Results` to `getDocuments` handler


alallema (approved):
LGTM! Great 🎉

PR #348, opened 2022-07-04
Add fields to get docs

brunoocasali:


alallema (commented):
LGTM! But why do we only provide fields as arguments and not all the [agitation parameters?

Issue #353, opened 2022-07-15
getDocuments fatal error


vitalijm:
**Description**
Line of code like: $client->index('movies')->getDocuments(['limit' => 2]); not returning array, but showing error:

**Expected behavior**
Return array of index items

**Current behavior**
<b>Fatal error</b>:  Uncaught TypeError: MeiliSearch\Endpoints\Indexes::getDocuments(): Argument #1 ($options) must be of type ?MeiliSearch\Contracts\DocumentsQuery, array given, called in /app/index.php on line 25 and defined in /app/vendor/meilisearch/meilisearch-php/src/Endpoints/Delegates/HandlesDocuments.php:22

When in HandlesDocuments.php I've changed:
```
    public function getDocuments(DocumentsQuery $options = null): DocumentsResults
    {
        $query = isset($options) ? $options->toArray() : [];
        $response = $this->http->get(self::PATH.'/'.$this->uid.'/documents', $query);

        return new DocumentsResults($response);
    }
```
to
```
    public function getDocuments($options): DocumentsResults
    {
        //$query = isset($options) ? $options->toArray() : [];
        $response = $this->http->get(self::PATH.'/'.$this->uid.'/documents', $options);

        return new DocumentsResults($response);
    }
```
then returns protect object with data
```
MeiliSearch\Contracts\DocumentsResults::__set_state(array(
   'data' => 
  array (
    0 => 
    array (
      'id' => 'some1',
      'title' => 'some1',
    ),
    1 => 
    array (
      'id' => 'some2',
      'title' => 'some2',
    ),
  ),
   'offset' => 0,
   'limit' => 2,
   'total' => 100,
))
```

**Environment (please complete the following information):**
- OS: DOCKER
- Meilisearch version: [v.0.28.0]
- meilisearch-php version: [v0.24.0]

norkunas:
As the error says, you must provide DocumentsQuery instead of array.

vitalijm:
I am trying to write php app code based on this documentation: https://docs.meilisearch.com/reference/api/documents.html#get-documents

norkunas:
Maybe outdated docs :) try to inspect the DocumentsQuery how to set the limit and provide it, should work :)

vitalijm:
> Maybe outdated docs :) try to inspect the DocumentsQuery how to set the limit and provide it, should work :)

Thanks. This ofcourse is possible (I think), but related to documentation for other prog languages, by logic there must passed options in simple array format.

[...]

PR #358, opened 2022-07-19
Use the correct notation for getting documents

brunoocasali:
Fix an issue pointed initially here https://github.com/meilisearch/meilisearch-php/issues/353

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -20,7 +20,7 @@ delete_an_index_1: |-
 get_one_document_1: |-
   $client->index('movies')->getDocument(25684, ['id', 'title', 'poster', 'release_date']);
 get_documents_1: |-
-  $client->index('movies')->getDocuments(['limit' => 2]);
+  $client->index('movies')->getDocuments((new DocumentsQuery())->setLimit(2));
 add_or_replace_documents_1: |-
   $client->index('movies')->addDocuments([
     [


tests/Search/SearchResultTest.php
```diff
@@ -229,4 +229,25 @@ public function testTransformFacetsDistributionMethod(): void
         $this->assertEquals(1, $response->getFacetDistribution()['genre']['FANTASY']);
         $this->assertEquals(1, $response->getFacetDistribution()['genre']['ADVENTURE']);
     }
+
+    public function testRonaldo(): void
```
alallema, line 233: Why this name of test?

src/Contracts/IndexesResults.php
```diff
@@ -20,7 +20,7 @@ public function __construct(array $params)
     }
 
     /**
-     * @return array<int, array>
+     * @return array<int, Indexes>
```
alallema, line 23: Php stan still doesn't seem happy with this, it is normal?

alallema (commented):
I'm not sure to fully understand the changes here

src/Contracts/IndexesResults.php
norkunas: Because the `Indexes` import is missing :)

norkunas: Also not sure why introducing it here if I've already made a PR for this fix? :D

tests/Search/SearchResultTest.php
brunoocasali: It was a mistake it shouldn't be commited, nevermind 🤦

alallema (approved):
LGTM! 🎉

brunoocasali:
bors merge

PR #373, opened 2022-08-17
Remove conditionals between steps in pre-release-tests.yml

brunoocasali:
I noticed we have only in the pre-release tests (GitHub action config file) some conditionals.

We do not need to add those since we already run without them in the ordinary tests action.


alallema (approved):
LGTM! Good catch

brunoocasali:
bors merge

PR #377, opened 2022-09-01
Ensure support to keys with wildcarded actions

brunoocasali:
Ensure support to keys with wildcarded actions.

- `actions` field during key creation now accepts wildcards on actions. For example, `indexes.*` provides rights to `indexes.create`, `indexes.get`,`indexes.delete`, `indexes.delete`, and `indexes.update`.


src/Search/SearchResult.php
```diff
@@ -60,8 +60,6 @@ public function __construct(array $body)
      * - transformHits (callable)
      *
      * The method does NOT trigger a new search.
-     *
-     * @return SearchResult
      */
```
bidoubiwa, line 63: I htink this changed is duplicated in #376

bidoubiwa (approved):
LGTM 🔥

PR #385, opened 2022-10-05
added type to http client property

b3none:
# Pull Request

## Related issue
Not related to an issue - should add contin

## What does this PR do?
It adds the type to the private class property `$http`.

## PR checklist
- [ ] Have you read the contributing guidelines?
- [ ] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -19,7 +19,7 @@ use MeiliSearch\Endpoints\TenantToken;
 use MeiliSearch\Endpoints\Version;
 use Psr\Http\Client\ClientInterface;
 use Psr\Http\Message\RequestFactoryInterface;
-use Http\Client as HttpClient;
+use \Http\Client as HttpClient;
 
 class Client
 {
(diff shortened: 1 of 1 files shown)


curquiza:
Hello @b3none 

Thanks for your PR 😊 
Since there is no issue, could you explain a little bit more about this change?
What is the purpose of it, I mean the value of changing this? If yes, is there any other part in the code base that would need this kind of change as well?

b3none:
> Hello @b3none
> 
> Thanks for your PR 😊 Since there is no issue, could you explain a little bit more about this change? What is the purpose of it, I mean the value of changing this? If yes, is there any other part in the code base that would need this kind of change as well?

@curquiza from what I’ve seen this is used everywhere, I think this one was missed because of the name class so I’ve switched to an alias.

The change makes the code more declarative, improve the speed and reliability of typehint generation as well as not allow anyone to set the $http class property to anything other than what it’s expecting.

curquiza:
Ok thank you!
However, all the tests are failing, can you fix it so that I can merge the PR?

b3none:
@curquiza try that

b3none:
Yeah I give up with this, i've already got my 4 hacktoberfest PRs 😂

PR #387, opened 2022-10-10
Allow to inject the StreamFactoryInterface explicitly

LeSuisse:
## What does this PR do?

This PR adds the possibility to inject `StreamFactoryInterface` to not rely on `php-http/discovery`.
The change is done in a way that does not break existing usages.

## PR checklist

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -19,6 +19,7 @@ use MeiliSearch\Endpoints\TenantToken;
 use MeiliSearch\Endpoints\Version;
 use Psr\Http\Client\ClientInterface;
 use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\StreamFactoryInterface;
 
 class Client
 {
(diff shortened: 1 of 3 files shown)


brunoocasali (changes requested):
Hello @LeSuisse!
 
Can you provide at least one use case of a test that covers this implementation?

LeSuisse:
Hello,

Yes I can.
We are using a custom StreamFactory to workaround an issue we have with Guzzle implementation see https://github.com/Enalean/tuleap/blob/97e900ad7812cb34bed5a68f73a3a8763080b8e4/src/common/Http/Factory/StreamFactory.php (well it looks like it was [fixed upstream since then](https://github.com/guzzle/psr7/commit/ea89a18a8a9d8abcc546da7a5de86791fdddb120)). Since this is a custom implementation it is not detected by `php-http/discovery`. This is not a big deal for the usage with `meilisearch-php` but I would prefer to have something consistent with no surprises.

I somehow missed the `tests/Http/ClientTest.php` class yesterday, I will add a unit test.

brunoocasali (approved):
Thanks a lot @LeSuisse <3


bors merge

brunoocasali:
_This message is sent automatically_

Thanks again for contributing to Meilisearch :heart:
If you are participating in [Hacktoberfest](https://blog.meilisearch.com/contribute-hacktoberfest-2022/), and you would like to receive some gift from Meilisearch too, please complete [this form](https://meilisearch.typeform.com/to/idt5qugY).

PR #419, opened 2022-11-21
Fix tasks specs with v0.30rc2

brunoocasali:
### Add and Update filters to the `tasks` resource.

- Added `setUids(array<int>)` filter to match by `Task.uid`.
- Added `setCanceledBy(array<int>)` filter to match by `Task.canceledBy`
- Added `setBeforeEnqueuedAt(\DateTime)` filter to match by `Task.beforeEnqueuedAt`
- Added `setAfterEnqueuedAt(\DateTime)` filter to match by `Task.AfterEnqueuedAt`
- Added `setBeforeStartedAt(\DateTime)` filter to match by `Task.beforeStartedAt`
- Added `setAfterStartedAt(\DateTime)` filter to match by `Task.AfterStartedAt`
- Added `setBeforeFinishedAt(\DateTime)` filter to match by `Task.beforeFinishedAt`
- Added `setAfterFinishedAt(\DateTime)` filter to match by `Task.AfterFinishedAt`
- ⚠️⚠️  Renamed `setUid(array)` to `setIndexUids(array)` filter to match by `Task.indexUid`
- ⚠️⚠️  Renamed `setStatus(array)` to `setStatuses(array<string>)` filter to match by `Task.status`

### DIFF
--- a/src/Delegates/TasksQueryTrait.php
+++ b/src/Delegates/TasksQueryTrait.php
@@ -1,0 +1,138 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MeiliSearch\Delegates;
+
+trait TasksQueryTrait
+{
+    private int $next;
+    private array $types;
+    private array $statuses;
+    private array $indexUids;
+    private array $uids;
+    private array $canceledBy;
+    private \DateTime $beforeEnqueuedAt;
+    private \DateTime $afterEnqueuedAt;
+    private \DateTime $beforeStartedAt;
+    private \DateTime $afterStartedAt;
+    private \DateTime $beforeFinishedAt;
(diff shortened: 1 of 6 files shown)


src/Delegates/TasksQueryTrait.php
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    public function setIndexUids(array $indexUids)
+    {
+        $this->indexUids = $indexUids;
+
+        return $this;
+    }
+
+    public function getIndexUids(): array
```
bidoubiwa, line 50: Why is there only one getter?

brunoocasali: There is no need to retrieve this data from other places, that's why :)

PR #422, opened 2022-11-23
Add `swapIndexes` method

brunoocasali:
Add a new method in the Client `client->swapIndexes(array $indexes)` to swap indexes. The array should follow this structure: `[['indexA', 'indexB'], ['indexC', 'indexD'], ...]`

### DIFF
--- a/src/Endpoints/Indexes.php
+++ b/src/Endpoints/Indexes.php
@@ -153,6 +153,9 @@ class Indexes extends Endpoint
         return $this->http->delete(self::PATH.'/'.$this->uid) ?? [];
     }
 
+    /**
+     * @param array{0: list<non-empty-string>, 1: list<non-empty-string>} $indexes
+     */
     public function swapIndexes(array $indexes): array
     {
         return $this->http->post('/swap-indexes', $indexes);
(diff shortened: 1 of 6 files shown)


src/Endpoints/Indexes.php
```diff
@@ -153,6 +153,11 @@ public function delete(): array
         return $this->http->delete(self::PATH.'/'.$this->uid) ?? [];
     }
 
+    public function swapIndexes(array $indexes): array
```
norkunas, line 156:
```suggestion
    /**
      * @param array{0: list<non-empty-string>, 1: list<non-empty-string>}
      */
    public function swapIndexes(array $indexes): array
```

Doc could help for phpstan :)

brunoocasali: Amazing, thanks @norkunas 🎉

bidoubiwa: >     * @param array{0: list<non-empty-string>, 1: list<non-empty-string>} $indexes

Not sure i understand the syntax, is it a type working with the following example?

```
[
  { indexes: ['a', 'b'] },
  { indexes: ['a', 'b'] }
]
```

norkunas: was on my phone, probably misunderstood from the tests.
then it should be:
```
@param list<array{indexes: list<non-empty-string>}>
```

bidoubiwa (commented):
Tests are failing

PR #431, opened 2022-12-16
Change casing of `MeiliSearch` to `Meilisearch`

mmachatschek:
# Pull Request

## Related issue
Fixes #<issue_number>

## What does this PR do?
This PR changes the casing from `MeiliSearch` to `Meilisearch`. 

To my surprise I discovered that PHP namespaces and classes are case insensitive and therefore recasing `Meilisearch` is possible. Therefore this change should not be breaking. I did some tests with the library and everything worked fine

See:

- https://www.php.net/manual/en/language.namespaces.rationale.php#:~:text=Note%3A%20Namespace%20names%20are%20case%2Dinsensitive.
- https://sebhastian.com/is-php-case-sensitive/

## PR checklist

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -423,7 +423,7 @@ getting_started_add_documents_md: |-
 
   require_once __DIR__ . '/vendor/autoload.php';
 
-  use MeiliSearch\Client;
+  use Meilisearch\Client;
 
   $client = new Client('http://localhost:7700');
 
--- a/.github/scripts/beta-docker-version.sh
+++ b/.github/scripts/beta-docker-version.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-meilisearch_php_version=$(grep 'public const VERSION =' ./src/MeiliSearch.php | cut -d ' ' -f 9 | tr -d "'" | tr -d ";")
+meilisearch_php_version=$(grep 'public const VERSION =' ./src/Meilisearch.php | cut -d ' ' -f 9 | tr -d "'" | tr -d ";")
 beta_feature=$(echo $meilisearch_php_version | sed -r 's/[0-9]+.[0-9]+.[0-9]+-meilisearch-//')
 beta_feature=$(echo $beta_feature | sed -r 's/-beta\.[0-9]*$//')
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,7 +104,7 @@ _[Read more about this](https://github.com/meilisearch/integration-guides/blob/m
 
 ⚠️ Before doing anything, make sure you got through the guide about [Releasing an Integration](https://github.com/meilisearch/integration-guides/blob/main/resources/integration-release.md).
 
-Make a PR modifying the file [`src/MeiliSearch.php`](/src/MeiliSearch.php) with the right version.
+Make a PR modifying the file [`src/Meilisearch.php`](/src/Meilisearch.php) with the right version.
 
 ```php
 const VERSION = 'X.X.X';
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ NB: you can also download Meilisearch from **Homebrew** or **APT** or even run i
 
 require_once __DIR__ . '/vendor/autoload.php';
 
-use MeiliSearch\Client;
+use Meilisearch\Client;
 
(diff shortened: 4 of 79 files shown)


brunoocasali (approved):
I share your surprise!

I made two suggestions, but in any case, is approved and thanks for that!

brunoocasali (dismissed):
I applied the changes myself 😋 

Meilisearch is made by people like you thanks a lot @mmachatschek ❤️

brunoocasali (approved):
bors merge

Nextra:
While PHP doesn't care about case for auto loading purposes, other tools in the ecosystem will, because the specific spelling ends up in `::class` strings:

```php
> Meilisearch\Client::class
= "Meilisearch\Client"
> MeiliSearch\Client::class
= "MeiliSearch\Client"

```

Anything that handles these strings directly - by using the class string as an array key for example - usually cares about casing. So while this is not a breaking change in the traditional sense, users might be required to use the old spelling to interact correctly with a third party dependency.

Example: Laravel's Scout registers a pre-configured Meilisearch Client as a singleton service:
https://github.com/laravel/scout/blob/4c91081ae2dbd59b1b72a6fae2ed1b9baa8dd0d2/src/ScoutServiceProvider.php#L26-L30

Observe how using this singleton will require the old spelling:

```php
> app()->make(Meilisearch\Client::class)
   # Illuminate\Contracts\Container\BindingResolutionException  Unresolvable dependency resolving [Parameter #0 [ <required> string $url ]] in class Meilisearch\Client.

> app()->make(MeiliSearch\Client::class)
= Meilisearch\Client {#4152}

```

If you keep the change, you might want to mention that this change occurred in the docs.

brunoocasali:
Hi @Nextra, this change will be appropriately handled on the Laravel side by @mmachatschek, who suggested the change :)

I've updated the release notes to include this relevant information. Thanks for sharing that!

mmachatschek:
@brunoocasali @Nextra on laravel scouts side, yes we export the client class from the container, but this is only for convenience.

@Nextra if you rely on the client you will need to stick to the old casing (from the container). with laravel v10 this will be changed and will definitely be mentioned in the upgrade guide.

Nextra:
I was aware of both the PHP loading behavior and the Laravel interaction (or those of similar caching methods), so I can handle it fine. Nice to see that Scout will have a major release to match where the change can be applied and make everything consistent again. Cheers.

mirceasoaica:
This is a major change and should not happen in a minor version change. 
Our production app crashed because of this change. On Mac OS both MeiliSearch and Meilisearch work, but on a linux environment they don't. Also Laravel Scout did not release an update for this change.

This should be a major version release! Not from 0.26.0 to 0.26.1.

mmachatschek:
@Nextra @mirceasoaica sorry for creating unexpected inconveniences on both of your sides. I created https://github.com/laravel/scout/pull/687 on the scout repo to support both of the casing syntaxes hope you guys can check it out and test it as well

jabstech:
+1, had to downgrade as well as our Linux server crashed with the same issue. 
Careful for upgrading to this version as it might break your linux environment until https://github.com/laravel/scout/pull/687 has been merged in.

n8man:
The changing of the namespace was handled fine by Composer, including on case-sensitive filesystems, but Composer remains case-sensitive for class names on case-sensitive filesystems... This is despite the fact that PHP is itself is case-**in**sensitive for class names and namespaces.

This definitely should not not have been part of a 0.26.0 -> 0.26.1 change. 1.0 would have been more appropriate.

n8man:
Is it reasonable at this point to keep the namespace change but revert the case change for the `MeiliSearch\MeiliSearch` class? I do have a PR that's mean to do that #441, but maybe that just creates more confusion.

My concern is for codebases that rely on meilisearch-php beyond laravel\scout.

brunoocasali:
The naming change was intentional, but the breaking wasn't since PHP handles it for us. In any case, I was naive to assume it wouldn't break in other communities, and I'm sorry about it.

Since the change is done, the breaking was not expected and many users had already updated their applications on the laravel side for example, the most straightforward solution is to release a new version.

https://packagist.org/packages/meilisearch/meilisearch-php

I released a new version, `v0.27.0`, without new additions just to mark it as breaking. 

@mmachatschek I think you can change [in the preparation PR](https://github.com/laravel/scout/pull/678) to use `v0.27` instead.

I'm sorry again for the inconvenience.

PR #441, opened 2023-01-05
Rename Meilisearch class to maintain autoload compatibility

n8man:
#431 Renamed the namespace from `MeiliSearch` to `Meilisearch`, which isn't a problem because the Composer config was also updated to allow PSR-4 matching of both cases.

The problem is that the class `MeiliSearch\MeiliSearch` was also renamed to `Meilisearch\Meilisearch`, not just the namespace. This breaks Composer autoloading because Composer class-name matching is case-sensitive.

This PR fixes the name of this class in order to maintain compatibility.

**If** the renaming of the that class is both intentional and permanent:
 - the README should be updated with a notice
 - the version number should be incremented by at least a new point release (0.27.x) instead of sub-point (0.26.**1** is not ideal to make a breaking change)

If you need a quick fix for laravel/scout or other projects, `composer require meilisearch/meilisearch-php "0.26.0"` should do the trick.

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -13,7 +13,7 @@ use Meilisearch\Exceptions\InvalidResponseBodyException;
 use Meilisearch\Exceptions\JsonDecodingException;
 use Meilisearch\Exceptions\JsonEncodingException;
 use Meilisearch\Http\Serialize\Json;
-use Meilisearch\Meilisearch;
+use Meilisearch\MeiliSearch;
 use Psr\Http\Client\ClientExceptionInterface;
 use Psr\Http\Client\ClientInterface;
 use Psr\Http\Client\NetworkExceptionInterface;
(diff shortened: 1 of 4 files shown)


driesvints:
I'd also very much opt to revert these changes as they're breaking. It's best that you only do this in a major release.

norkunas:
[class_alias](https://www.php.net/class_alias) could be used to provide compatibility

brunoocasali:
Hi @n8man, I'm sorry that renaming brings so many headaches for everyone in the community.

I would like to keep the current naming `Meilisearch\Meilisearch`.

I released it as a patch because PHP deals with those names case-insensitively without breaking, and it was not supposed to be breaking.

So, I'll release a new version, `0.27.0`, and mark `0.26.1` as **deprecated on packagist**. I'll also put a notice in the release changelog for both versions explaining the new version.

Since many users are already using the new version `0.26.1` with laravel/scout @driesvints don't you think it will bring more breakings? But now for the people who already fixed the environment?

driesvints:
Yeah a workaround now might be better

brunoocasali:
> Yeah a workaround now might be better

Released a new package version `v0.27.0`. Unfortunately, I didn't find a way to inform the users automatically by packagist, as I said before because this is a basic feature I can do in different stacks. 

I'll also close this PR @n8man, thanks for the help it is very much appreciated.

PR #448, opened 2023-01-18
Add PHP 8.2 to CI

norkunas:
# Pull Request

## Related issue
Fixes #<issue_number>

## What does this PR do?
- Adds PHP 8.2 to CI tests

## PR checklist


tests/Endpoints/IndexTest.php
```diff
@@ -192,7 +192,10 @@ public function testGetTasks(): void
         $results = array_unique($allIndexUids);
         $expected = [$this->index->getUid(), 'other-index'];
 
-        $this->assertSame(sort($results), sort($expected));
+        sort($results);
+        sort($expected);
+
+        $this->assertSame($results, $expected);
```
norkunas, line 198: `sort` always returns true, so phpstan complained about that: `Call to method PHPUnit\Framework\Assert::assertSame() with true and true will always evaluate to true.`

.github/workflows/tests.yml
```diff
@@ -18,7 +18,8 @@ jobs:
       - name: Install PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 8.0
+          php-version: 8.1
+          coverage: none
```
norkunas, line 22: By default when non setting `coverage: none` xdebug is installed, so this just avoids

.github/workflows/tests.yml
```diff
@@ -32,7 +33,8 @@ jobs:
       - name: Install PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 8.0
+          php-version: 8.2
+          coverage: none
```
norkunas, line 37: Same here

.github/workflows/tests.yml
```diff
@@ -45,14 +47,15 @@ jobs:
...
-        php-versions: ['7.4', '8.0', '8.1']
+        php-versions: ['7.4', '8.0', '8.1', '8.2']
     name: integration-tests (PHP ${{ matrix.php-versions }})
     steps:
     - uses: actions/checkout@v1
     - name: Install PHP
       uses: shivammathur/setup-php@v2
       with:
         php-version: ${{ matrix.php-versions }}
+        coverage: none
```
norkunas, line 58: Later we could introduce coverage if needed, but currently also disabled to avoid installing xdebug

tests/Endpoints/IndexTest.php
norkunas: And voila, now tests fails because the expectations were wrong :D
```
1) Tests\Endpoints\IndexTest::testGetTasks
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
     0 => 'index_[29](https://github.com/meilisearch/meilisearch-php/actions/runs/3946584981/jobs/6754483260#step:7:30)2b37af8be3c1639bd46d3d8b08a16f'
+    1 => 'other-index'
 )
```

norkunas: Not sure what this code is supposed to test so waiting for feedback

brunoocasali: Hi @norkunas thanks for pointing this out.

The idea was to test the ability to compose a `getTasks` request from an index instance where I could also use the value defined inside the instance.

In this case, we want to get all the tasks from both `$this->index` and `other-index`. And we just want to ensure the presence of both indexUids in the response. They don't need to respect the order in the array btw.

When I implemented this test I didn't realize that `sort()` affects the value instead of creating a new array 🤦.

norkunas: There was a bug :)

norkunas:
All green, if you want I can split a bugfix into separate PR

brunoocasali (approved):
bors merge

Thanks for your contribution @norkunas 🤘

PR #454, opened 2023-02-01
Remove raw methods & Fix query classes

brunoocasali:
- Add `canceledBy` and remove `next` from toArray in `TasksQuery`
- Correct `DeleteTasksQuery` allowed attributes: removed `next`.
- Correct `CancelTasksQuery` allowed attributes: removed `next` and `canceledBy`.
- Correct `TasksQuery` allowed attributes: removed `next`. 
- Remove non-generic methods from `TasksQueryTrait` 
- Remove `deleteAllIndexes`, `getAllRawIndexes`.
- Rename `getAllIndexes` to `getIndexes`.
- Refactor `tearDown` after removing `deleteAllIndexes`

### DIFF
--- a/src/Contracts/CancelTasksQuery.php
+++ b/src/Contracts/CancelTasksQuery.php
@@ -9,23 +9,4 @@ use Meilisearch\Delegates\TasksQueryTrait;
 class CancelTasksQuery
 {
     use TasksQueryTrait;
-
-    private array $canceledBy;
-
-    public function setCanceledBy(array $canceledBy)
-    {
-        $this->canceledBy = $canceledBy;
-
-        return $this;
-    }
-
-    public function toArray(): array
-    {
-        return array_filter(
-            array_merge(
-                $this->baseArray(),
-                ['canceledBy' => $this->formatArray($this->canceledBy ?? null)]
-            ), function ($item) { return null != $item || is_numeric($item); }
-        );
-    }
 }
--- a/src/Contracts/CancelTasksQuery.php
+++ b/src/Contracts/CancelTasksQuery.php
@@ -9,4 +9,23 @@ use Meilisearch\Delegates\TasksQueryTrait;
 class CancelTasksQuery
 {
     use TasksQueryTrait;
+
+    private array $canceledBy;
+
+    public function setCanceledBy(array $canceledBy)
+    {
+        $this->canceledBy = $canceledBy;
+
+        return $this;
+    }
+
+    public function toArray(): array
+    {
+        return array_filter(
+            array_merge(
+                $this->baseArray(),
+                ['canceledBy' => $this->formatArray($this->canceledBy ?? null)]
+            ), function ($item) { return null != $item || is_numeric($item); }
+        );
+    }
 }
--- a/src/Delegates/HandlesIndex.php
+++ b/src/Delegates/HandlesIndex.php
@@ -10,14 +10,9 @@ use Meilisearch\Endpoints\Indexes;
 
 trait HandlesIndex
 {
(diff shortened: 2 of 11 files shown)


src/Contracts/CancelTasksQuery.php
```diff
@@ -9,4 +9,23 @@
...
+
+    public function toArray(): array
+    {
+        return array_filter(
+            array_merge(
+                $this->baseArray(),
+                ['canceledBy' => $this->formatArray($this->canceledBy ?? null)]
+            ), function ($item) { return null != $item || is_numeric($item); }
+        );
+    }
```
alallema, line 30: I thought that this method was removed from `TasksQueryTrait` specifically because it was not used in `CancelTasksQuery` no?

src/Delegates/HandlesIndex.php
```diff
@@ -10,16 +10,11 @@
...
+    public function getIndexes(IndexesQuery $options = null): IndexesResults
     {
         return $this->index->all($options ?? null);
     }
 
-    public function getAllRawIndexes(IndexesQuery $options = null): array
-    {
-        return $this->index->allRaw($options ?? []);
-    }
-
```
alallema, line 22: I'm not sure we should remove this method even if alright the user may not need it. But if we do, we should remove `allRaw` method too no? Or it is used somewhere else?

src/Delegates/TasksQueryTrait.php
```diff
@@ -110,8 +94,15 @@ public function setAfterFinishedAt(\DateTime $date)
 
     public function toArray(): array
     {
-        return array_filter([
-            'next' => $this->next ?? null,
+        return array_filter(
+            $this->baseArray(),
+            function ($item) { return null != $item || is_numeric($item); }
+        );
+    }
```
alallema, line 101: Nice!

src/Contracts/CancelTasksQuery.php
brunoocasali: Not really. This method is required for every `Query` class. The problem is not correct to access stuff from the `Trait`.  So now, the idea is that the Trait generate most of the content (the shared part), and I just add the missing parts, like `canceledBy` in this case.

src/Delegates/HandlesIndex.php
brunoocasali: Yeah, that's taking my sleep away... I'm not comfortable changing most of the "raw" methods, because, internally the Indexes class uses a "very strange pattern to me".

alallema: Me too, I removed some of them in some SDK, but I wasn't sure either ... Maybe we should have a discussion about it with @meilisearch/integration-team? WDYT?

src/Contracts/CancelTasksQuery.php
alallema: Yes, but technically you can't cancel a task that has already been canceled so if there is no point in filtering them via `canceledBy` as they have already been canceled yet no?
Or I missed something

src/Delegates/HandlesIndex.php
brunoocasali: Yes, let's do it!

alallema (dismissed):
LGTM! 🌮

brunoocasali:
@alallema I just merged in the wrong way, I should have merged this branch first. I'll ask for your review again here.

brunoocasali:
bors merge

alallema (dismissed):

alallema (approved):

brunoocasali:
bors merge

PR #457, opened 2023-02-02
Make createdAt and updatedAt DateTime types not string

brunoocasali:
- Replace the `string` type by `?DateTime` in `Endpoints\Indexes::createdAt` and `Endpoints\Indexes::updatedAt`

### DIFF
--- a/src/Endpoints/Indexes.php
+++ b/src/Endpoints/Indexes.php
@@ -27,8 +27,8 @@ class Indexes extends Endpoint
 
     private ?string $uid;
     private ?string $primaryKey;
-    private ?string $createdAt;
-    private ?string $updatedAt;
+    private ?\DateTime $createdAt;
+    private ?\DateTime $updatedAt;
     private Tasks $tasks;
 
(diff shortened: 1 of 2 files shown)


norkunas:
Why not start to use `DateTimeImmutable`? :) One day DateTime could be deprecated https://www.reddit.com/r/PHP/comments/w7v4pb/deprecating_the_mutable_datetime_class/

brunoocasali:
> Why not start to use `DateTimeImmutable`? :) One day DateTime could be deprecated

Hi @norkunas, I appreciate the idea and agree with it, but I think it is too soon to make a change like that. Let's wait for more info on the PHP core team.

norkunas:
> > Why not start to use `DateTimeImmutable`? :) One day DateTime could be deprecated
> 
> Hi @norkunas, I appreciate the idea and agree with it, but I think it is too soon to make a change like that. Let's wait for more info on the PHP core team.

Okay but why is your preference is to use `DateTime` which has no upsides over `DateTimeImmutable`?

alallema (approved):
LGTM! 🌯

brunoocasali:
bors merge

brunoocasali:
bors cancel

PR #480, opened 2023-03-21
Add CSV delimiter to csv document handler methods

brunoocasali:
- Add `?string $delimiter` to `updateDocumentsCsv`, `addDocumentsCsv` and also `addDocumentsCsvInBatches` and `updateDocumentsCsvInBatches`.

### DIFF
--- a/tests/Endpoints/DocumentsTest.php
+++ b/tests/Endpoints/DocumentsTest.php
@@ -84,6 +84,26 @@ final class DocumentsTest extends TestCase
         $this->assertCount(20, $response);
     }
 
+    public function testAddDocumentsCsvWithCustomSeparator(): void
+    {
+        $index = $this->client->index('documentCsvWithCustomSeparator');
+
+        $csv = file_get_contents('./tests/datasets/songs-custom-separator.csv', true);
+
+        $promise = $index->addDocumentsCsv($csv, null, '|');
+
+        $this->assertIsValidPromise($promise);
+
+        $update = $index->waitForTask($promise['taskUid']);
+
+        $this->assertEquals($update['status'], 'succeeded');
+        $this->assertEquals($update['details']['receivedDocuments'], 6);
+
+        $documents = $index->getDocuments()->getResults();
+        $this->assertEquals('Teenage Neon Jungle', $documents[4]['album']);
+        $this->assertEquals('631152000', $documents[5]['released-timestamp']);
+    }
+
     public function testAddDocumentsJson(): void
     {
         $index = $this->client->index('documentJson');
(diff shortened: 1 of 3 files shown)


tests/Endpoints/DocumentsTest.php
```diff
@@ -280,6 +302,34 @@ public function testAddDocumentsCsvInBatches(): void
         $this->assertSame($total, $response->getTotal());
     }
 
+    public function testAddDocumentsCsvInBatchesWithDelimiter(): void
```
norkunas, line 305: would it fail with empty delimiter? if yes, maybe better to document then as `non-empty-string|null` ?

tests/Endpoints/DocumentsTest.php
```diff
@@ -610,6 +687,34 @@ public function testUpdateDocumentsCsvInBatches(): void
...
+
+        $index->expects($matcher)
+              ->method('updateDocumentsCsv')
+              ->willReturnCallback(function (string $param) use ($matcher): void {
+                  // withConsecutive has no replacement https://github.com/sebastianbergmann/phpunit/issues/4026
+                  // @phpstan-ignore-next-line
+                  match ($matcher->numberOfInvocations()) {
+                      1 => $this->assertEquals($param, ["id;title\n888221515;Young folks", null, ';']),
+                      2 => $this->assertEquals($param, ["id;title\n235115704;Mister Klein", null, ';']),
+                  };
```
norkunas, line 711:
```suggestion
                      2 => $this->assertEquals($param, ["id;title\n235115704;Mister Klein", null, ';']),
                      default => self::fail(),
                  };
```

tests/Endpoints/DocumentsTest.php
norkunas: With this phpdoc users could know what it would fail even before calling the api

alallema (approved):
So great! ✨ Remind me that I didn't check the.`CsvInBatches` methods
LGTM!

tests/Endpoints/DocumentsTest.php
brunoocasali: Yes, thanks @alallema I'll also add your suggestion @norkunas :)

tests/Endpoints/DocumentsTest.php
```diff
@@ -84,6 +84,28 @@ public function testAddDocumentsCsv(): void
         $this->assertCount(20, $response);
     }
 
+    public function testAddDocumentsCsvWithCustomSeparator(): void
+    {
+        $index = $this->client->index('documentCsvWithCustomSeparator');
+
+        $fileCsv = fopen('./tests/datasets/songs-custom-separator.csv', 'r');
+        $documentCsv = fread($fileCsv, filesize('./tests/datasets/songs-custom-separator.csv'));
+        fclose($fileCsv);
```
norkunas, line 93: why not just `file_get_contents`?

brunoocasali: Good idea! I'm not much aware of all the PHP features, so I'm glad you suggested that :)

PR #481, opened 2023-03-21
Add `multiSearch`

brunoocasali:
- Add a new class `SearchQuery` which will handle future search requests
- Add a new method in `$client` called `multiSearch()` which will take an array of `SearchQuery`.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -8,6 +8,7 @@ use Meilisearch\Contracts\Http as HttpContract;
 use Meilisearch\Endpoints\Delegates\HandlesDumps;
 use Meilisearch\Endpoints\Delegates\HandlesIndex;
 use Meilisearch\Endpoints\Delegates\HandlesKeys;
+use Meilisearch\Endpoints\Delegates\HandlesMultiSearch;
 use Meilisearch\Endpoints\Delegates\HandlesSystem;
 use Meilisearch\Endpoints\Delegates\HandlesTasks;
 use Meilisearch\Endpoints\Dumps;
@@ -30,6 +31,7 @@ class Client
     use HandlesTasks;
     use HandlesKeys;
     use HandlesSystem;
+    use HandlesMultiSearch;
 
     private HttpContract $http;
     private Indexes $index;
--- a/src/Contracts/SearchQuery.php
+++ b/src/Contracts/SearchQuery.php
@@ -173,6 +173,6 @@ class SearchQuery
             'limit' => $this->limit ?? null,
             'hitsPerPage' => $this->hitsPerPage ?? null,
             'page' => $this->page ?? null,
-        ], function ($item) { return null != $item || is_numeric($item); });
+        ], function ($item) { return null !== $item || is_numeric($item); });
     }
 }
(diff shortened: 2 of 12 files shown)


src/Client.php
```diff
@@ -30,6 +31,7 @@ class Client
     use HandlesTasks;
     use HandlesKeys;
     use HandlesSystem;
+    use HandlesMultiSearch;
```
norkunas, line 34: Just interesting - what's the decision to use so many traits that are used only once here and that just hides the code?

src/Endpoints/Delegates/HandlesMultiSearch.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Meilisearch\Endpoints\Delegates;
+
+trait HandlesMultiSearch
+{
+    public function multiSearch(array $queries = [])
```
norkunas, line 9:
```suggestion
{
    /**
     * @param list<SearchQuery> $queries
     */
    public function multiSearch(array $queries = [])
```

people with static analysis will thank you :)

norkunas: also `array<SearchQuery>` would also work

src/Contracts/SearchQuery.php
```diff
@@ -0,0 +1,178 @@
...
+            'highlightPostTag' => $this->highlightPostTag ?? null,
+            'facets' => $this->facets ?? null,
+            'showMatchesPosition' => $this->showMatchesPosition ?? null,
+            'sort' => $this->sort ?? null,
+            'matchingStrategy' => $this->matchingStrategy ?? null,
+            'offset' => $this->offset ?? null,
+            'limit' => $this->limit ?? null,
+            'hitsPerPage' => $this->hitsPerPage ?? null,
+            'page' => $this->page ?? null,
+        ], function ($item) { return null != $item || is_numeric($item); });
```
norkunas, line 176:
```suggestion
        ], function ($item) { return null !== $item || is_numeric($item); });
```

src/Client.php
brunoocasali: I'm following the adopted pattern, but in fact, I agree because the `Client` class would be humongous if all those methods were defined directly there.

src/Endpoints/Delegates/HandlesMultiSearch.php
brunoocasali: Do you know how can I run this locally without phpstorm?

src/Client.php
bidoubiwa: Hey @norkunas, 
@brunoocasali left on holliday so I'll be releasing this PR. I agree with you as well, nonetheless I suggest we open an issue on this subject for a possible refacto of that section and merge this one as it is. What do you think?

bidoubiwa (approved):

src/Endpoints/Delegates/HandlesMultiSearch.php
norkunas: @brunoocasali sorry missed this comment, not sure what you meant :)

src/Client.php
norkunas: hi @bidoubiwa, i don't mind :)

PR #488, opened 2023-03-30
Use wildcards in tests

brunoocasali:
Just use a wildcard somewhere to ensure the suite will still work after it.

### DIFF
--- a/tests/Endpoints/TenantTokenTest.php
+++ b/tests/Endpoints/TenantTokenTest.php
@@ -26,7 +26,7 @@ final class TenantTokenTest extends TestCase
         $this->key = $this->client->createKey([
             'description' => 'tenant token key',
             'actions' => ['*'],
-            'indexes' => ['*'],
+            'indexes' => ['tenant*'],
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       - ./:/home/package
 
   meilisearch:
-    image: getmeili/meilisearch:latest
+    image: getmeili/meilisearch:v1.1.0-rc.2
```
bidoubiwa, line 19: Do we really want to change this?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM php:8.1
+FROM php:7.4
```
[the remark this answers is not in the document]
brunoocasali: the curse of using `git add .` 🤣

docker-compose.yml
brunoocasali: the curse of using `git add .` 🤣

bidoubiwa (approved):
LGTM 🔥

PR #502, opened 2023-05-10
Add patch to prevent breaking in laravel/scout

RoccoHoward:
# Add patch to prevent breaking in laravel/scout

## What does this PR do?

Fixes Undefined array key "totalHits" in laravel/scout [10.x],

A patch was made to provide the `nbHits` but from v10 of Scout this is now `totalHits` to reflect the change from Meilisearch v1.

See the related code in scout v10, 
https://github.com/laravel/scout/blob/877bce500c7d1cc467176dea9bc0c87640c270fb/src/Engines/MeilisearchEngine.php#L319

```
   /**
     * Get the total count from a raw result returned by the engine.
     *
     * @param  mixed  $results
     * @return int
     */
    public function getTotalCount($results)
    {
        return $results['totalHits'];
    }
```

## PR checklist

### DIFF
--- a/src/Endpoints/Indexes.php
+++ b/src/Endpoints/Indexes.php
@@ -206,7 +206,7 @@ class Indexes extends Endpoint
         // patch to prevent breaking in laravel/scout getTotalCount method,
         // affects only Meilisearch >= v0.28.0.
         if (isset($result['estimatedTotalHits'])) {
-            $result['nbHits'] = $result['estimatedTotalHits'];
+            $result['totalHits'] = $result['nbHits'] = $result['estimatedTotalHits'];
         }
 
         return $result;


norkunas:
I don't understand why low level library should know something about laravel/scout? Why you don't fix this in it?

RoccoHoward:
I'm just going off the fact that this patch exists to deal with a Scout issue.

See existing comment:
```
// patch to prevent breaking in laravel/scout getTotalCount method,
// affects only Meilisearch >= v0.28.0.
```

Including  @driesvints - this bug fixes a pagination issue in Scout v10 - which may need to be fixed at Scout's end long term?

norkunas:
I'd say that if laravel/scout needs to support multiple meilisearch versions for BC then it should be done on it instead of here

PR #506, opened 2023-05-23
Update Data.php

### LINKED ISSUE #505: add @return mixed to remove deprecation warning
[deprecation] User Deprecated: Method "ArrayAccess::offsetGet()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Meilisearch\Contracts\Data" now to avoid errors or add an explicit @return annotation to suppress this message. ["exception" => ErrorException { …}]

Is there any downside to adding @return mixed to this method?  Alas, php7.4 doesn't support return type mixed.

```php
    /**
     * {@inheritDoc}
     */
    #[\ReturnTypeWillChange]
    public function offsetGet($offset)
    {
        if (isset($this->data[$offset])) {
            return $this->data[$offset];
        }

        return null;
    }
```

tacman:
add @return mixed to get rid of deprecation message, #505

# Pull Request

## Related issue
Fixes #505

## What does this PR do?
- gets rid of deprecation message

### DIFF
--- a/src/Contracts/Data.php
+++ b/src/Contracts/Data.php
@@ -39,6 +39,7 @@ class Data implements \ArrayAccess, \Countable, \IteratorAggregate
 
     /**
      * {@inheritDoc}
+     * @return mixed 
      */
     #[\ReturnTypeWillChange]
     public function offsetGet($offset)


brunoocasali:
Hey @tacman since the problem was bigger than we imagined I had to fix it #508 here, but thank you very much for taking the time for doing it anyway, I'll mention your username in the release changelogs!

PR #509, opened 2023-05-23
Add ability to use deleteDocuments with filters after v1.2

brunoocasali:
Change the `deleteDocuments()` function where it is now possible to send in the argument a key `filter`, which will enable the ability to filter and delete documents.

Ex: `deleteDocuments(['filter' => ['price > 10']]);`

⚠️ Remind to add the attributes used to filter into the filterable attributes before calling the delete method `deleteDocuments`.
⚠️ The current behavior stays the same.
⚠️ If you want to use the new behavior upgrade to Meilisearch v1.2 or newer.


src/Endpoints/Delegates/HandlesDocuments.php
```diff
@@ -139,9 +141,19 @@ public function deleteDocument($documentId): array
...
+        try {
+            if (\array_key_exists('filter', $options) && $options['filter']) {
+                return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete', $options);
+            } else {
+                // backwards compatibility:
+                // expect to be a array to send alongside as $documents_ids.
+                return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete-batch', $options);
+            }
+        } catch (InvalidResponseBodyException $e) {
+            throw ApiException::rethrowWithHint($e, __FUNCTION__);
```
norkunas, line 155: this should be covered with test

src/Exceptions/ApiException.php
```diff
@@ -85,4 +87,9 @@ private function getErrorLinkFromHttpBody(): ?string
 
         return null;
     }
+
+    public static function rethrowWithHint(\Exception $e, string $methodName)
+    {
+        return new \Exception(str_replace('_METHOD_', $methodName, self::HINT_MESSAGE), 0, $e);
```
stloyd, line 93: Instead of `str_replace` you should use `sprintf()` with just a `%s` placeholder.

src/Endpoints/Delegates/HandlesDocuments.php
```diff
@@ -139,9 +141,19 @@ public function deleteDocument($documentId): array
...
     {
-        return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete-batch', $documents);
+        try {
+            if (\array_key_exists('filter', $options) && $options['filter']) {
+                return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete', $options);
+            } else {
+                // backwards compatibility:
+                // expect to be a array to send alongside as $documents_ids.
+                return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete-batch', $options);
+            }
```
stloyd, line 153:
```suggestion
            if (\array_key_exists('filter', $options) && $options['filter']) {
                return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete', $options);
            }

            // backwards compatibility:
            // expect to be a array to send alongside as $documents_ids.
            return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete-batch', $options);
```

tests/Exceptions/ApiExceptionTest.php
```diff
@@ -59,4 +59,23 @@ public function testExceptionWithNoHttpBody(): void
...
         }
     }
+
+    public function testRethrowWithHintException(): void
+    {
+        try {
+            $e = new \Exception('Any error message that caused the root problem should be shown');
+
+            throw ApiException::rethrowWithHint($e, 'deleteDocuments');
+        } catch (\Exception $apiException) {
```
norkunas, line 69: `expectException` should be used instead of catching

brunoocasali: Actually, this is impossible because I need to assert if the previous exception error message is preserved. Unless you know how to do it.

Also, your commend made me realize I don't need to throw, I could just ensure the messages are how I want.

src/Endpoints/Delegates/HandlesDocuments.php
```diff
@@ -139,9 +141,19 @@ public function deleteDocument($documentId): array
...
+        try {
+            if (\array_key_exists('filter', $options) && $options['filter']) {
+                return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete', $options);
+            }
+
+            // backwards compatibility:
+            // expect to be a array to send alongside as $documents_ids.
+            return $this->http->post(self::PATH.'/'.$this->uid.'/documents/delete-batch', $options);
+        } catch (InvalidResponseBodyException $e) {
+            throw ApiException::rethrowWithHint($e, __FUNCTION__);
```
alallema, line 155: I don't know exactly what it is `ApiException`
If the request is sent to v.1.1 the return will be a 405 `Method Not Allowed` so it would be a `ComunicationError` instead.
Isn't there an error plus something more general? Because the day we have a new field that is not a filter in a future version, it will be followed by a possible `ApiException`.
Maybe it would be good to time do it now but it's up to you

alallema (commented):
Well Done 🎉

src/Endpoints/Delegates/HandlesDocuments.php
brunoocasali: I tried locally by using a different version in the docker-compose, and that was the error I got. To improve this I think I could catch any error and give always give the hint.

brunoocasali: `ApiException` is just a class I used to put the method `rethrowWithHint` on it. Nothing special there, but I choose it because it is the closest class I have to a `BaseError` class or something

src/Exceptions/ApiException.php
```diff
@@ -85,4 +87,9 @@ private function getErrorLinkFromHttpBody(): ?string
 
         return null;
     }
+
+    public static function rethrowWithHint(\Exception $e, string $methodName)
+    {
+        return new \Exception(sprintf(self::HINT_MESSAGE, $methodName), 0, $e);
```
norkunas, line 93: why instead of throwing ApiException as you define the static constructor in it, you throw generic exception?

src/Endpoints/Delegates/HandlesDocuments.php
alallema: I have reverified in all my SDKs and a 405 returns a `MeilisearchAPIError` in my SDKs not a `CommunicationError`.
My bad

src/Exceptions/ApiException.php
brunoocasali: Good question. I decided because it has the simplest constructor. All Meilisearch error classes require a `ResponseInterface $response` to be sent alongside it. To make that happen, I would have to deal with the `$e` and make some validations before rethrowing too much work for no apparent gain.

PR #510, opened 2023-05-24
Add get documents

brunoocasali:
Introduces a new method, `setFilter` to `DocumentsQuery`, which will add the ability to `getDocuments` to find the documents easily without using the `search` method. 

⚠️ If you want to use the new feature, ensure you're using Meilisearch v1.2.0 or newer.

### DIFF
--- a/src/Contracts/DocumentsQuery.php
+++ b/src/Contracts/DocumentsQuery.php
@@ -32,6 +32,13 @@ class DocumentsQuery
         return $this;
     }
 
+    /**
+     * Sets the filter for the DocumentsQuery.
+     *
+     * @param list<non-empty-string|list<non-empty-string>> $filter a filter expression written as an array of strings
+     *
+     * @return DocumentsQuery the updated DocumentsQuery instance
+     */
     public function setFilter(array $filter): DocumentsQuery
     {
         $this->filter = $filter;
@@ -39,6 +46,11 @@ class DocumentsQuery
         return $this;
     }
 
+    /**
+     * Checks if the $filter attribute has been set.
+     *
+     * @return bool true when filter contains at least an empty array
+     */
     public function hasFilter(): bool
     {
         return isset($this->filter);
--- a/src/Contracts/DocumentsQuery.php
+++ b/src/Contracts/DocumentsQuery.php
@@ -49,7 +49,7 @@ class DocumentsQuery
     /**
      * Checks if the $filter attribute has been set.
      *
-     * @return bool true when filter contains at least an empty array
+     * @return bool true when filter contains at least a non-empty array
      */
     public function hasFilter(): bool
     {
--- a/src/Contracts/DocumentsQuery.php
+++ b/src/Contracts/DocumentsQuery.php
@@ -9,6 +9,7 @@ class DocumentsQuery
     private int $offset;
     private int $limit;
     private array $fields;
+    private array $filter;
 
     public function setOffset(int $offset): DocumentsQuery
     {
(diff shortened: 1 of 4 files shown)


src/Contracts/DocumentsQuery.php
```diff
@@ -31,11 +32,24 @@ public function setFields(array $fields): DocumentsQuery
         return $this;
     }
 
+    public function setFilter(array $filter): DocumentsQuery
```
norkunas, line 35: should add phpdoc what is `$filter` :)

src/Endpoints/Delegates/HandlesDocuments.php
```diff
@@ -22,10 +22,20 @@ public function getDocument($documentId, array $fields = null)
...
+
+            if ($options->hasFilter()) {
+                $response = $this->http->post(self::PATH.'/'.$this->uid.'/documents/fetch', $query);
+            } else {
+                $response = $this->http->get(self::PATH.'/'.$this->uid.'/documents', $query);
+            }
 
-        return new DocumentsResults($response);
+            return new DocumentsResults($response);
+        } catch (\Exception $e) {
```
norkunas, line 36: this should probably be a more specialized exception from http than a generic one

brunoocasali: Yes, I agree with you! I'm awaiting an issue that @alallema will create in the integration-guides. We will have to check it across every SDK. 
For example, we don't have a "BaseMeilisearchException" class that we can use across the package to check if anything happens. Today all of our exceptions are extending Exception 😢.

src/Contracts/DocumentsQuery.php
norkunas: I don't. At least for new methods could be good to add :)

brunoocasali: I've added the phpdoc. Can you check it?

norkunas: Looking at the tests it's only list<string> or list<non-empty-string>

Or are there any cases with nested arrays?

brunoocasali: They are valid, I just don't need to test all the cases across every SDK (we are trying to avoid that as much as possible).

You can have this: `setFilter([['id > 10'], ['price < 250']])` which will translate to: "id > 10 AND price < 250"

norkunas: Okay so `list<non-empty-string|list<non-empty-string>>`

brunoocasali: done! ✅

src/Contracts/DocumentsQuery.php
```diff
@@ -31,11 +32,36 @@ public function setFields(array $fields): DocumentsQuery
...
+    {
+        $this->filter = $filter;
+
+        return $this;
+    }
+
+    /**
+     * Checks if the $filter attribute has been set.
+     *
+     * @return bool true when filter contains at least an empty array
```
norkunas, line 52:
```suggestion
     * @return bool true when filter contains at least a non-empty array
```

alallema (approved):
Well Done 🔥 
LGTM! 🎉 

Just a question, what does the modification of the pre-release GitHub action do in this PR?

src/Endpoints/Delegates/HandlesDocuments.php
alallema: It's out:
- [About the error handler](https://github.com/meilisearch/integration-guides/issues/267)
- [About the hint](https://github.com/meilisearch/integration-guides/issues/268)

brunoocasali:
> Just a question, what does the modification of the pre-release GitHub action do in this PR?

The branching tree!

PR #511, opened 2023-05-24
Bump meilisearch v1.2.0 try pre release rc ci

brunoocasali:
Fix PHP HTTP libraries deprecations in the Pre-release CI structure.

Also, for some reason, only the `kriswallsmith/buzz` was breaking because of the way I set up the `ResponseInterface` mock.


alallema (approved):
LGTM! 🎉

PR #514, opened 2023-05-30
[DX] Improve quality of Client class & related

stloyd:
# Pull Request

## Related issue
Fixes n/a

## What does this PR do?

The changes in this PR aim to improve the quality of the code:
- moving property definition to traits that use them,
- adding type-hints when possible,
- adding array-shapes into the param definition when possible,
- simplifying some code

## PR checklist

### DIFF
--- a/src/Endpoints/Delegates/HandlesSystem.php
+++ b/src/Endpoints/Delegates/HandlesSystem.php
@@ -4,8 +4,18 @@ declare(strict_types=1);
 
 namespace Meilisearch\Endpoints\Delegates;
 
+use Meilisearch\Endpoints\Health;
+use Meilisearch\Endpoints\Stats;
+use Meilisearch\Endpoints\TenantToken;
+use Meilisearch\Endpoints\Version;
+
 trait HandlesSystem
 {
+    protected Health $health;
+    protected Version $version;
+    protected TenantToken $tenantToken;
+    protected Stats $stats;
+
     public function health(): ?array
     {
         return $this->health->show();
@@ -32,14 +42,14 @@ trait HandlesSystem
         return $this->stats->show();
     }
 
-    public function generateTenantToken(string $apiKeyUid, $searchRules, ?array $options = []): string
+    public function generateTenantToken(string $apiKeyUid, $searchRules, array $options = []): string
     {
         return $this->tenantToken->generateTenantToken($apiKeyUid, $searchRules, $options);
     }
 
-    public function swapIndexes(array $indexes)
+    public function swapIndexes(array $indexes): array
     {
-        $options = array_map(function ($data) { return ['indexes' => $data]; }, $indexes);
+        $options = array_map(static fn ($data) => ['indexes' => $data], $indexes);
 
         return $this->index->swapIndexes($options);
     }
--- a/src/Endpoints/TenantToken.php
+++ b/src/Endpoints/TenantToken.php
@@ -11,17 +11,20 @@ use Meilisearch\Http\Serialize\Json;
 
 class TenantToken extends Endpoint
 {
-    private function base64url_encode($data)
+    private function base64url_encode(string $data): string
     {
         return rtrim(strtr(base64_encode($data), '+/', '-_'), '=');
     }
 
-    private function validateTenantTokenArguments($searchRules, ?array $options = []): void
+    /**
+     * @param array{apiKey?: ?string, expiresAt?: ?\DateTimeInterface} $options
+     */
+    private function validateTenantTokenArguments($searchRules, array $options = []): void
     {
-        if (!\array_key_exists('apiKey', $options) || '' == $options['apiKey'] || \strlen($options['apiKey']) <= 8) {
+        if (!\array_key_exists('apiKey', $options) || '' === $options['apiKey'] || \strlen($options['apiKey']) <= 8) {
             throw InvalidArgumentException::emptyArgument('api key');
         }
-        if ((!\is_array($searchRules) && !\is_object($searchRules)) || null == $searchRules) {
+        if (!\is_array($searchRules) && !\is_object($searchRules)) {
             throw InvalidArgumentException::emptyArgument('search rules');
         }
         if (\array_key_exists('expiresAt', $options) && new \DateTime() > $options['expiresAt']) {
@@ -35,10 +38,12 @@ class TenantToken extends Endpoint
      * The $options parameter is an array, and the following keys are accepted:
      * - apiKey: The API key parent of the token. If you leave it empty the client API Key will be used.
      * - expiresAt: A DateTime when the key will expire. Note that if an expiresAt value is included it should be in UTC time.
+     *
+     * @param array{apiKey?: ?string, expiresAt?: ?\DateTimeInterface} $options
      */
-    public function generateTenantToken(string $uid, $searchRules, ?array $options = []): string
+    public function generateTenantToken(string $uid, $searchRules, array $options = []): string
     {
-        if (!\array_key_exists('apiKey', $options) || '' == $options['apiKey']) {
+        if (!\array_key_exists('apiKey', $options) || '' === $options['apiKey']) {
             $options['apiKey'] = $this->apiKey;
         }
 
@@ -51,7 +56,7 @@ class TenantToken extends Endpoint
         $header = [
             'typ' => 'JWT',
             'alg' => 'HS256',
-          ];
+        ];
 
         // Add the required fields to the payload
         $payload = [];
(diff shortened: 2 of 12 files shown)


brunoocasali:
Hi @stloyd thanks a lot for your contribution. Quick question: do these improvements introduce any breaking change?

Also, I have other PRs in my queue to review so I'll probably take a look at your code near the end of the week, ok?

Good to have you here!

stloyd:
Hey, no problem about the later review.

About changes, those are mostly to prevent later bugs rather than introduce them, there is only one case when those changes could introduce BC break if someone would extend classes from this SDK and try to overwrite the methods.

From my perspective I would even suggest (or do) a PR that will lock all the classes with the `final` keyword, to prevent the BC breaks, and allow easier maintenance and a safer Developer Experience in the end. 

Refs: https://ocramius.github.io/blog/when-to-declare-classes-final/

brunoocasali:
As far as I know, the main integrations we have will not be impacted even after adding the `final` classes, but I will want to have the opinion of our stars.

📡 Signal to @mmachatschek (scout) and @norkunas (symfony).

stloyd:
To not pollute this PR with discussion about `final`'s I have added new task about it: https://github.com/meilisearch/meilisearch-php/issues/518

mmachatschek:
Concerning the `final` topic. I hear a lot of negatives about this (in the Laravel community) to not use this feature, even though it looks like a good idea. Users that need to extend classes marked as final will need to wait for upstream or have to fork libs themselves if they run into issues.
Also there are workarounds to extend classes even if they are marked as final so there is no 100% guarantee that the class marked as final will be in the final state.

I haven't looked much into the changes on this PR but @brunoocasali takes care to test this thoroughly. I will try to test theses changes with a laravel scout project as soon as @brunoocasali has done his first review 👍

src/Endpoints/Delegates/HandlesSystem.php
```diff
@@ -32,14 +42,14 @@ public function stats(): array
         return $this->stats->show();
     }
 
-    public function generateTenantToken(string $apiKeyUid, $searchRules, ?array $options = []): string
+    public function generateTenantToken(string $apiKeyUid, $searchRules, array $options = []): string
```
brunoocasali, line 45: This is a breaking change, no? Since we are now making a requirement to call with the array instead of an optional.

src/Endpoints/Delegates/HandlesSystem.php
```diff
@@ -32,14 +42,14 @@ public function stats(): array
...
     {
         return $this->tenantToken->generateTenantToken($apiKeyUid, $searchRules, $options);
     }
 
-    public function swapIndexes(array $indexes)
+    public function swapIndexes(array $indexes): array
     {
-        $options = array_map(function ($data) { return ['indexes' => $data]; }, $indexes);
+        $options = array_map(static fn ($data) => ['indexes' => $data], $indexes);
```
brunoocasali, line 52: Nicee! 🎉

src/Endpoints/Delegates/HandlesSystem.php
```diff
@@ -4,8 +4,18 @@
...
 namespace Meilisearch\Endpoints\Delegates;
 
+use Meilisearch\Endpoints\Health;
+use Meilisearch\Endpoints\Stats;
+use Meilisearch\Endpoints\TenantToken;
+use Meilisearch\Endpoints\Version;
+
 trait HandlesSystem
 {
+    protected Health $health;
```
brunoocasali, line 14: To understand it entirely, since the traits are a mechanism for code reuse, we are now defining the properties that are used inside of the trait internally instead of simply waiting for the class that will `use Trait` to define them, right?

If the statement above is right, why defining this in such a way is even possible? Do you know any history behind this language decision?

brunoocasali (commented):
I usually take the review time to improve my skills in the languages and learn with the experts, so if you don't mind I made some comments in that purpose, but also there is one about the `?array` change that may not be time to do it, since in my view it is breaking!

Thanks a lot for your time, and I will wait your answer, thanks!! 🎉

src/Http/Client.php
```diff
@@ -27,13 +27,13 @@ class Client implements Http
...
     private StreamFactoryInterface $streamFactory;
+    /** @var array<string,string> */
     private array $headers;
-    private ?string $apiKey;
     private string $baseUrl;
     private Json $json;
 
     /**
-     * Client constructor.
+     * @param array<int, string> $clientAgents
```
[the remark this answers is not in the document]
stloyd: They are type-hinted so IDEs will handle this as well as static analyze tools.

src/Exceptions/InvalidArgumentException.php
```diff
@@ -24,7 +24,7 @@ public static function emptyArgument(string $argumentName): self
         );
     }
 
-    public static function dateIsExpired(\DateTime $date): self
+    public static function dateIsExpired(\DateTimeInterface $date): self
```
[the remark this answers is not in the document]
stloyd: Yes exactly.

norkunas:
> As far as I know, the main integrations we have will not be impacted even after adding the `final` classes, but I will want to have the opinion of our stars.
> 
> satellite Signal to @mmachatschek (scout) and @norkunas (symfony).

for me it looks like a good decision to make them final, as composition should be always preferred vs inheritance

src/Endpoints/TenantToken.php
```diff
@@ -11,20 +11,23 @@
...
     }
 
-    private function validateTenantTokenArguments($searchRules, ?array $options = []): void
+    /**
+     * @param array{apiKey?: ?string, expiresAt?: ?\DateTimeInterface} $options
+     */
+    private function validateTenantTokenArguments($searchRules, array $options = []): void
     {
-        if (!\array_key_exists('apiKey', $options) || '' == $options['apiKey'] || \strlen($options['apiKey']) <= 8) {
+        if (!isset($options['apiKey']) || ('' === $options['apiKey'] || \strlen($options['apiKey']) <= 8)) {
```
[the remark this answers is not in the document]
stloyd: There were three fails here:
- allowing nullable when we expect an array with particular values to be set,
- `==` which is not strict comparison, but flaky one, which will try to convert on fly values to string, which is dangerous,
- `array_key_exists()` which allows passing a `null` value, which would only work if the above one is still old (flaky `==`)

src/Http/Client.php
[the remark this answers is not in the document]
norkunas: maybe just `list<non-empty-string>` ?

src/Endpoints/Delegates/HandlesSystem.php
stloyd: In a really rare case it can be a BC break, but it's optional already, as it has the default value `$options = []` which will pass `[]` when the developer doesn't pass anything on its own.

But here it leads to another case described here: https://github.com/meilisearch/meilisearch-php/pull/514/files#r1221466672 which would mean that now the code is less fragile and more strict for things that are required to be set, or how they should be set.

src/Endpoints/Delegates/HandlesSystem.php
stloyd: This change is to make DX better, while trait is for code sharing, they would fail badly when you don't declare required properties for them in the class that uses them (previous version of code), while with `protected` (which means shared within inheritance but still not publicly visible) means that developer can use methods by just adding those traits with not further changes.

In fact, those traits in this SDK are overkill IMHO, cause the code here is not really so abstract to be shared as much as it seems, cause those traits in fact are definitions of endpoints, which will not really change over time, yet they will not be shared outside of their scopes (you will not call 2-3 endpoints from 2-3 traits in same SDK method).

brunoocasali: Perfect, thanks for the explanation!

brunoocasali (approved):
bors merge

src/Http/Client.php
```diff
@@ -44,15 +44,14 @@ public function __construct(
...
-        $this->apiKey = $apiKey;
         $this->http = $httpClient ?? Psr18ClientDiscovery::find();
         $this->requestFactory = $reqFactory ?? Psr17FactoryDiscovery::findRequestFactory();
         $this->streamFactory = $streamFactory ?? Psr17FactoryDiscovery::findStreamFactory();
         $this->headers = array_filter([
             'User-Agent' => implode(';', array_merge($clientAgents, [Meilisearch::qualifiedVersion()])),
         ]);
-        if (null != $this->apiKey) {
-            $this->headers['Authorization'] = sprintf('Bearer %s', $this->apiKey);
+        if (null !== $apiKey) {
```
mmachatschek, line 53: @brunoocasali unfortunately this made some troubles for environment keys that are passed in as empty string as mentioned here: 

https://github.com/laravel/scout/issues/769

norkunas: Imho we should add phpdoc `@param non-empty-string|null $apiKey`

mmachatschek: @norkunas I'm posting the issue description now for tracking

PR #523, opened 2023-06-06
Ppunk documents fetch fix #

### LINKED ISSUE #522: Get documents with filters: Invalid value type at `.fields`: expected an array, but found a string
Implode here
https://github.com/meilisearch/meilisearch-php/blob/main/src/Contracts/DocumentsQuery.php#L65

Post request with the fields as string here:
https://github.com/meilisearch/meilisearch-php/blob/main/src/Endpoints/Delegates/HandlesDocuments.php#L30

I'm gonna contribute the fix.

panzer-punk:
Fixes #522 

Fixed behavior for fields if new documents/fetch endpoint used.

### DIFF
--- a/src/Contracts/DocumentsQuery.php
+++ b/src/Contracts/DocumentsQuery.php
@@ -64,7 +64,7 @@ class DocumentsQuery
      *
      * @return array|string|null
      */
-    private function fields(): mixed
+    private function fields()
     {
         if (!isset($this->fields)) {
             return null;
--- a/src/Contracts/DocumentsQuery.php
+++ b/src/Contracts/DocumentsQuery.php
@@ -56,13 +56,33 @@ class DocumentsQuery
         return isset($this->filter);
     }
 
+    /**
+     * Prepares fields for request
+     *
+     * @return array
+     */
+    private function fields(): string|array|null
+    {
+        if (! isset($this->fields)) {
+            return null;
+        }
+
+        if ($this->hasFilter()) {
+            return $this->fields;
+        }
+
+        else {
+            return implode(",", $this->fields);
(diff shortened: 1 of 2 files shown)


src/Contracts/DocumentsQuery.php
```diff
@@ -56,13 +56,34 @@ public function hasFilter(): bool
...
 
+    /**
+     * Prepares fields for request
+     * Fix for 1.2 document/fetch.
+     *
+     * @see https://github.com/meilisearch/meilisearch-php/issues/522
+     *
+     * @return array|string|null
+     */
+    private function fields(): mixed
```
brunoocasali, line 67: The `mixed` broke the implementation it is only introduced in PHP 8 right? So we can't use that, because we still support php 7.4

```suggestion
    private function fields()
```

brunoocasali (changes requested):
Hi @panzer-punk thanks a lot for your bug fix, I just pointed out something that broke the code.

I'm not sure if there is a replacement, please let me know.

src/Contracts/DocumentsQuery.php
panzer-punk: @brunoocasali yep, my bad. I'll fix it

panzer-punk:
@brunoocasali removed mixed return type

brunoocasali:
bors merge

brunoocasali:
bors merge

panzer-punk:
@brunoocasali I'm not sure what's going on, I just tried to sync fork and I might've broken some checks. bors said there is some integration failure with guzzle 7 and php 7.4. Sorry

PR #543, opened 2023-07-10
Add support to attributesToSearchOn

brunoocasali:
Add support for receiving a new key, `attributesToSearchOn` (which is an array of field names), that will enable running searches over a subset of `searchableAttributes` without modifying Meilisearch’s index settings.

### DIFF
--- a/src/Contracts/SearchQuery.php
+++ b/src/Contracts/SearchQuery.php
@@ -26,6 +26,7 @@ class SearchQuery
     private ?int $hitsPerPage;
     private ?int $page;
     private ?array $vector;
+    private ?array $attributesToSearchOn;
 
     public function setQuery(string $q): SearchQuery
     {
@@ -169,6 +170,13 @@ class SearchQuery
         return $this;
     }
 
+    public function setAttributesToSearchOn(array $attributesToSearchOn): SearchQuery
+    {
+        $this->attributesToSearchOn = $attributesToSearchOn;
+
+        return $this;
+    }
+
     public function toArray(): array
     {
         return array_filter([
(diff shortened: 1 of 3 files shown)


brunoocasali:
@alallema the tests are breaking because the branch is not pointing to the bump branch.

src/Contracts/SearchQuery.php
```diff
@@ -26,6 +26,7 @@ class SearchQuery
     private ?int $hitsPerPage;
     private ?int $page;
     private ?array $vector;
+    private ?array $attributesToSearchOn;
```
norkunas, line 29: I'd instead initialize props as null to avoid null coalescing in `toArray`

src/Contracts/SearchQuery.php
```diff
@@ -169,6 +170,13 @@ public function setVector(array $vector): SearchQuery
         return $this;
     }
 
+    public function setAttributesToSearchOn(array $attributesToSearchOn): SearchQuery
```
norkunas, line 173:
```suggestion
    /**
     * @param list<non-empty-string> $attributesToSearchOn
     */
    public function setAttributesToSearchOn(array $attributesToSearchOn): SearchQuery
```

src/Contracts/SearchQuery.php
brunoocasali: You mean: `private ?array $attributesToSearchOn = null;`?

norkunas: Yup :)

norkunas: But then would better to do this everywhere if needed

brunoocasali: I made it for the `attributesToSearchOn` and I'll raise an issue to do it for the others :)

brunoocasali: Done: https://github.com/meilisearch/meilisearch-php/issues/547

PR #544, opened 2023-07-10
Add support to `showRankingScore`

brunoocasali:
This feature aims to return ranking details for each document to understand and tweak the score of the documents more efficiently.

Ensure the SDKs can handle the new search parameter `showRankingScore`. Also, ensure the SDK can handle the `_rankingScore` attribute in the matched `hits`.


alallema (approved):
LGTM! 🌮

PR #546, opened 2023-07-11
Added ability to init SearchQuery props on construction

Enchiridion:
# Pull Request

## What does this PR do?
This lets you init the `SearchQuery` props when the object is constructed.  I think the PHP library is the only one that doesn't yet let you do this or use a simple array instead.  I created this because I was using the `search()` method but needed to switch to `multiSearch()`, but found it works completely differently, this speeds up using it.

Previous way:
```
$data = (new SearchQuery())
   ->setIndexUid('books')
   ->setQuery('butler')
   ->setSort(['author:desc']);
```

Additional way: 
```
$data = new SearchQuery([
    'indexUid' => 'books',
    'q'        => 'butler',
    'sort'     => ['author:desc'],
]);
```

Note 1: My change currently does not pass the test due to the use of variable property access.  However I think it's safe because all the props are simple and typed.  If you think that's ok, I can have phpstan ignore that line.

```
 ------ -----------------------------------------------------------------------
  Line   src/Contracts/SearchQuery.php
 ------ -----------------------------------------------------------------------
  33     Variable property access on $this(Meilisearch\Contracts\SearchQuery).
 ------ -----------------------------------------------------------------------
```

Note 2: The initial code I based this on from `main` currently does not pass all the phpstan tests:

```
 ------ ------------------------------------------------------------------------------
  Line   tests/Endpoints/DocumentsTest.php
 ------ ------------------------------------------------------------------------------
  324    Call to an undefined method
         PHPUnit\Framework\MockObject\Rule\InvokedCount::numberOfInvocations().
  805    Call to an undefined method
         PHPUnit\Framework\MockObject\Rule\InvokedAtLeastOnce::numberOfInvocations().
 ------ ------------------------------------------------------------------------------
```

## PR checklist

### DIFF
--- a/src/Contracts/SearchQuery.php
+++ b/src/Contracts/SearchQuery.php
@@ -26,6 +26,35 @@ class SearchQuery
     private ?int $hitsPerPage;
     private ?int $page;
 
+    // @phpstan-ignore-next-line
+    public function __construct(
+        string $indexUid = null,
+        string $q = null,
+        array $filter = null,
+        array $attributesToRetrieve = null,
+        array $attributesToCrop = null,
+        int $cropLength = null,
+        array $attributesToHighlight = null,
+        string $cropMarker = null,
+        string $highlightPreTag = null,
+        string $highlightPostTag = null,
+        array $facets = null,
+        bool $showMatchesPosition = null,
+        array $sort = null,
+        string $matchingStrategy = null,
+        int $offset = null,
+        int $limit = null,
+        int $hitsPerPage = null,
+        int $page = null,
+    ) {
+        foreach (get_defined_vars() as $name => $value) {
+            if (!\is_null($value) && property_exists($this, $name)) {
+                // @phpstan-ignore-next-line
+                $this->$name = $value;
+            }
+        }
+    }
+
     public function setQuery(string $q): SearchQuery
     {
         $this->q = $q;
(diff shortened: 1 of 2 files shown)


Enchiridion:
I added in the linter ignore for the Variable property access line.

norkunas:
Is it really worth? With array you don't have autocompletion, because the structure is undefined

norkunas:
> I added in the linter ignore for the Variable property access line.

I'd document the structure instead of ignoring

Enchiridion:
I've updated the PR so all the properties are in the constructor.  With PHP8+ you can specify the args using named arguments along with array spreading to get the same effect of setting all those properties quickly and easily using an array.

src/Contracts/SearchQuery.php
```diff
@@ -26,6 +26,35 @@ class SearchQuery
...
+        array $facets = null,
+        bool $showMatchesPosition = null,
+        array $sort = null,
+        string $matchingStrategy = null,
+        int $offset = null,
+        int $limit = null,
+        int $hitsPerPage = null,
+        int $page = null,
+    ) {
+        foreach (get_defined_vars() as $name => $value) {
```
norkunas, line 50: Set the vars manually then the phpstan won't complain and in future it could be refactored to CPP

norkunas (changes requested):

curquiza:
@Enchiridion do you still plan to work on this PR? 😊

PR #551, opened 2023-07-12
Add getTotal() in the TasksResults type

brunoocasali:
Add `total` to `TasksResults`.


alallema (approved):
LGTM!

src/Contracts/TasksResults.php
```diff
@@ -9,6 +9,7 @@ class TasksResults extends Data
     private int $next;
     private int $limit;
     private int $from;
+    private int $total;
```
norkunas, line 12: if it cannot be <0, then maybe better to typehint `@var int<0, max>` ?

PR #585, opened 2023-10-11
Fix passing empty masterkey

### LINKED ISSUE #584: Issue when upgrading from < 1.3 to >= 1.3 with api keys set to empty string
**Description**

PR #514 cleaned up some stuff around the HTTP Client and made some improvements. Unfortunately the PR introduced an breaking change where previously an empty string `''` would not add authorization headers to subsequent api requests to the Meilisearch instance.

e.g. PHP env files at Laravel, this env var would be an empty string:

```
MEILISEARCH_KEY=
```

**Expected behavior**

This code should not add an authorization header to subsequent Meilisearch requests.

```
use Meilisearch\Client;
$meilisearchClient = new Client(
  'http://127.0.0.1:7700',
  '',
);
```

**Current behavior**

The above code adds auth headers to Meilisearch requests.

**Screenshots or Logs**

https://github.com/laravel/scout/issues/769

**Environment (please complete the following information):**
- meilisearch-php version: >= 1.3.0

mmachatschek:
# Pull Request

## Related issue
Fixes #584

## What does this PR do?
- Fix the issue by also checking if the provided masterkey is an empty string

## PR checklist

### DIFF
--- a/tests/Http/ClientTest.php
+++ b/tests/Http/ClientTest.php
@@ -239,11 +239,8 @@ class ClientTest extends TestCase
         $requestStub->expects($this->once())
             ->method('withAddedHeader')
             ->willReturnCallback(function ($name, $value) use ($requestStub) {
-                if ('User-Agent' === $name) {
-                    $this->assertSame(Meilisearch::qualifiedVersion(), $value);
-                } else {
-                    $this->fail('There should be nothing else than User-Agent header');
-                }
+                $this->assertSame('User-Agent', $name);
+                $this->assertSame(Meilisearch::qualifiedVersion(), $value);
 
                 return $requestStub;
             });
--- a/tests/Http/ClientTest.php
+++ b/tests/Http/ClientTest.php
@@ -239,8 +239,10 @@ class ClientTest extends TestCase
         $requestStub->expects($this->once())
             ->method('withAddedHeader')
             ->willReturnCallback(function ($name, $value) use ($requestStub) {
-                if ('Authorization' === $name) {
-                    $this->assertSame('Bearer ', $value);
+                if ('User-Agent' === $name) {
+                    $this->assertSame(Meilisearch::qualifiedVersion(), $value);
+                } else {
+                    $this->fail('There should be nothing else than User-Agent header');
                 }
 
                 return $requestStub;
--- a/tests/Http/ClientTest.php
+++ b/tests/Http/ClientTest.php
@@ -248,7 +248,7 @@ class ClientTest extends TestCase
             ->method('createRequest')
             ->willReturn($requestStub);
 
-        $client = new Client('http://localhost:7070', '', $httpClient, $reqFactory);
+        $client = new \Meilisearch\Client('http://localhost:7070', '', $httpClient, $reqFactory);
 
         $this->assertTrue($client->isHealthy());
     }
--- a/tests/Http/ClientTest.php
+++ b/tests/Http/ClientTest.php
@@ -231,6 +231,29 @@ class ClientTest extends TestCase
         $this->assertTrue($client->isHealthy());
     }
 
+    public function testEmptyMasterkeyRemovesAuthHeader(): void
+    {
+        $httpClient = $this->createHttpClientMock(200, '{}');
+        $reqFactory = $this->createMock(RequestFactoryInterface::class);
+        $requestStub = $this->createMock(RequestInterface::class);
+        $requestStub->expects($this->exactly(1))
+            ->method('withAddedHeader')
+            ->willReturnCallback(function ($name, $value) use ($requestStub) {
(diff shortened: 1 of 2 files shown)


tests/Http/ClientTest.php
```diff
@@ -231,6 +231,29 @@ public function testClientHasDefaultUserAgent(): void
         $this->assertTrue($client->isHealthy());
     }
 
+    public function testEmptyMasterkeyRemovesAuthHeader(): void
+    {
+        $httpClient = $this->createHttpClientMock(200, '{}');
+        $reqFactory = $this->createMock(RequestFactoryInterface::class);
+        $requestStub = $this->createMock(RequestInterface::class);
+        $requestStub->expects($this->exactly(1))
```
norkunas, line 239:
```suggestion
        $requestStub->expects($this->once())
```

mmachatschek:
@curquiza I didn't revert to the previous behaviour, instead I opted for an check if the user provided an empty string. Empty string can not be passed via authorization anyway so I think this should be safe

norkunas (dismissed):
On new major release I'd leave as is documenting that non-empty-string or null are accepted, so maybe a note for future could be added

tests/Http/ClientTest.php
```diff
@@ -231,6 +231,29 @@ public function testClientHasDefaultUserAgent(): void
...
+    {
+        $httpClient = $this->createHttpClientMock(200, '{}');
+        $reqFactory = $this->createMock(RequestFactoryInterface::class);
+        $requestStub = $this->createMock(RequestInterface::class);
+        $requestStub->expects($this->once())
+            ->method('withAddedHeader')
+            ->willReturnCallback(function ($name, $value) use ($requestStub) {
+                if ('Authorization' === $name) {
+                    $this->assertSame('Bearer ', $value);
+                }
```
mmachatschek, line 244: this actually tests the wrong thing. I'll update the test now

mmachatschek: @norkunas thanks for the suggestion, didn't see the comment while providing a fix 😅 didn't write PHP tests in a while.

Can you check the latest commit?

https://github.com/meilisearch/meilisearch-php/pull/585/commits/dd2d9db35bc5c42d9a298f84ef884f341488a843

mmachatschek:
> On new major release I'd leave as is documenting that non-empty-string or null are accepted, so maybe a note for future could be added

@norkunas I don't quite understand. What do you exactly mean?

tests/Http/ClientTest.php
norkunas: looks good :)

norkunas:
> > On new major release I'd leave as is documenting that non-empty-string or null are accepted, so maybe a note for future could be added
> 
> @norkunas I don't quite understand. What do you exactly mean?

I mean to add a note about https://github.com/meilisearch/meilisearch-php/pull/514#discussion_r1348705607 to not forget it in future :)

tests/Http/ClientTest.php
norkunas: on the second thought, as there is `$this->once()`, the else case will never be reached I think.
so this would be enough:
```
        $requestStub->expects($this->once())
            ->method('withAddedHeader')
            ->willReturnCallback(function ($name, $value) use ($requestStub) {
                $this->assertSame('User-Agent', $name);
                $this->assertSame(Meilisearch::qualifiedVersion(), $value);
```

tests/Http/ClientTest.php
```diff
@@ -231,6 +231,28 @@ public function testClientHasDefaultUserAgent(): void
...
+                $this->assertSame('User-Agent', $name);
+                $this->assertSame(Meilisearch::qualifiedVersion(), $value);
+
+                return $requestStub;
+            });
+        $reqFactory->expects($this->once())
+            ->method('createRequest')
+            ->willReturn($requestStub);
+
+        $client = new \Meilisearch\Client('http://localhost:7070', '', $httpClient, $reqFactory);
```
norkunas, line 251:
```suggestion
        $client = new Client('http://localhost:7070', '', $httpClient, $reqFactory);
```

brunoocasali (approved):
LGTM!

norkunas:
Maybe we should add deprecation?

curquiza:
bors try

curquiza:
@mmachatschek @norkunas I think the tests are still failing 😕

tests/Http/ClientTest.php
mmachatschek: @norkunas this made the test fail because I specifically was using the `\Meilisearch\Client::class` instead of the `\Meilisearch\Http\Client::class` 😅

norkunas:
Strange to have two Client classes :)

norkunas (approved):

norkunas:
bors merge

PR #589, opened 2023-10-26
Add CI to update library version

curquiza:
I cannot test it directly before merging it sorry -> this is how `workflow_dispatch` works with GitHub Actions

@mmachatschek and @norkunas I pinged both of you to be aware of the tool, but I only need one review 😊

### DIFF
--- a/.github/workflows/update-version.yml
+++ b/.github/workflows/update-version.yml
@@ -1,0 +1,37 @@
+name: Update library version
+
+on:
+  workflow_dispatch:
+    inputs:
+      new_version:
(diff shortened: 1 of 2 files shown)


norkunas:
Maybe we could integrate https://github.com/Roave/BackwardCompatibilityCheck

curquiza:
@norkunas good idea!
I would rather we do this in another PR however 😊

norkunas (approved):

mmachatschek (approved):
I don't have a strong opinion on which Github Action tool to use. but the changes LGTM

curquiza:
bors merge

PR #607, opened 2024-01-10
Update XML dist files

brunoocasali:
It has been a while since I don't touch the PHP repo, and I had to run `composer update` which raised a warning about having outdated XML files.

I don't think this PR is warmful because I guess it is related to a new PHPUnit version.

And here we are!

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,20 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.1/phpunit.xsd"
-         bootstrap="vendor/autoload.php"
-         convertErrorsToExceptions="true"
-         convertNoticesToExceptions="true"
-         convertWarningsToExceptions="true"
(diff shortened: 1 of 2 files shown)


phpunit.xml.dist
```diff
@@ -1,20 +1,12 @@
...
-    <testsuites>
-        <testsuite name="Tests">
-            <directory suffix="Test.php">./tests</directory>
-        </testsuite>
-    </testsuites>
-    <php>
-        <env name="MEILISEARCH_URL" value="http://localhost:7700"/>
-        <env name="MEILISEARCH_API_KEY" value="masterKey"/>
-    </php>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.5/phpunit.xsd" bootstrap="vendor/autoload.php" stopOnFailure="false" cacheResult="false" colors="true" cacheDirectory=".phpunit.cache">
```
norkunas, line 2: Maybe reference local schema that is shipped with phpunit,so you can open it in IDE?

norkunas: By the way, probably schemas are different on 9.5/10 so two of them should be shipped and selected which to run on which version, but you should check,maybe I'm wrong

PR #608, opened 2024-01-15
Add hybrid search for v1.6.0

brunoocasali:
Adds hybrid search to PHP SDK. 

There is a commented test that should be reverted as soon as the related issue is fixed on the engine.

### DIFF
--- a/src/Endpoints/Delegates/HandlesSettings.php
+++ b/src/Endpoints/Delegates/HandlesSettings.php
@@ -361,4 +361,21 @@ trait HandlesSettings
     {
         return $this->http->delete(self::PATH.'/'.$this->uid.'/settings/proximity-precision');
     }
+
+    // Settings - Experimental: Embedders (hybrid search)
+
+    public function getEmbedders(): array
+    {
+        return $this->http->get(self::PATH.'/'.$this->uid.'/settings/embedders');
+    }
+
+    public function updateEmbedders(array $embedders): array
+    {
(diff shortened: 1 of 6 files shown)


src/Endpoints/Delegates/HandlesSettings.php
```diff
@@ -361,4 +361,21 @@ public function resetProximityPrecision(): array
...
     }
+
+    // Settings - Experimental: Embedders (hybrid search)
+
+    public function getEmbedders(): array
+    {
+        return $this->http->get(self::PATH.'/'.$this->uid.'/settings/embedders');
+    }
+
+    public function updateEmbedders(array $embedders): array
```
norkunas, line 372: `$embedders` structure should be defined

curquiza: @norkunas we don't have the time to do it for this PR, can we open a PR for the community to update it?

norkunas: yup, but I don't find anywhere docs about this :)

norkunas: @curquiza from https://www.meilisearch.com/docs/learn/experimental/vector_search and https://github.com/meilisearch/meilisearch/releases/tag/v1.6.0 I understand that structure should be like: `array<non-empty-string, array{source: non-empty-string, apiKey?: non-empty-string, model?: non-empty-string, documentTemplate?: non-empty-string, dimensions: positive-int}>` can you confirm?

PR #645, opened 2024-06-06
Add get similar document method

### LINKED ISSUE #642: [v1.9.0] Get similar documents
Related to https://github.com/meilisearch/integration-guides/issues/301

Introduction of a new route to get similar documents

Usage: https://meilisearch.notion.site/Get-similar-documents-usage-540919ca755c4da0b7cdee273db3f290?pvs=74

### TODO

- [ ] Add a new method named `searchSimilarDocuments()` having 1 parameter. This parameter is an object correspond to the body you should pass to the `POST /similar` route. Do not manage error, Meilisearch will handle them if a field is missing in the body.
Do NOT implement the route for `GET /similar`, we will add it in a second iteration if requested.
- [ ] Add tests. Here is the [basic test](https://github.com/meilisearch/meilisearch-php/pull/645#discussion_r1632879832)

⚠️ Make PRs pointing to `bump-meilisearch-v1.9.0` and NOT `main`. Please do 1 PR for all of these changes, and not several.

the-sinner:
# Pull Request

## Related issue
Fixes #642 

## What does this PR do?
- Add get similar document method

## PR checklist

### DIFF
--- a/src/Contracts/SimilarDocumentsQuery.php
+++ b/src/Contracts/SimilarDocumentsQuery.php
@@ -1,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Meilisearch\Contracts;
+
+class SimilarDocumentsQuery
+{
+    private int|string $id;
+    private ?int $offset = null;
+    private ?int $limit = null;
+    private ?string $embedder = null;
+    private ?array $attributesToRetrieve = null;
+    private ?bool $showRankingScore = null;
+    private ?bool $showRankingScoreDetails = null;
+    private ?array $filter = null;
+
+    public function setId(string|int $id): SimilarDocumentsQuery
+    {
+        $this->id = $id;
+
+        return $this;
+    }
+
+    public function setOffset(?int $offset): SimilarDocumentsQuery
+    {
+        $this->offset = $offset;
+
+        return $this;
+    }
+
+    public function setLimit(?int $limit): SimilarDocumentsQuery
+    {
+        $this->limit = $limit;
+
+        return $this;
+    }
+
+    public function setFilter(array $filter): SimilarDocumentsQuery
+    {
+        $this->filter = $filter;
+
+        return $this;
+    }
+
+    public function setEmbedder(string $embedder): SimilarDocumentsQuery
+    {
+        $this->embedder = $embedder;
+
+        return $this;
+    }
+
+    public function setAttributesToRetrieve(array $attributesToRetrieve): SimilarDocumentsQuery
+    {
+        $this->attributesToRetrieve = $attributesToRetrieve;
+
+        return $this;
+    }
+
+    public function setShowRankingScore(?bool $showRankingScore): SimilarDocumentsQuery
+    {
+        $this->showRankingScore = $showRankingScore;
+
+        return $this;
+    }
+
+    public function setShowRankingScoreDetails(?bool $showRankingScoreDetails): SimilarDocumentsQuery
+    {
+        $this->showRankingScoreDetails = $showRankingScoreDetails;
+
+        return $this;
+    }
+
+    public function toArray(): array
+    {
+        return array_filter([
+            'id' => $this->id,
+            'offset' => $this->offset,
+            'limit' => $this->limit,
+            'filter' => $this->filter,
+            'embedder' => $this->embedder,
+            'attributesToRetrieve' => $this->attributesToRetrieve,
+            'showRankingScore' => $this->showRankingScore,
+            'showRankingScoreDetails' => $this->showRankingScoreDetails,
+        ], function ($item) { return null !== $item; });
+    }
+}
--- a/src/Contracts/SimilarDocumentsQuery.php
+++ b/src/Contracts/SimilarDocumentsQuery.php
@@ -6,7 +6,10 @@ namespace Meilisearch\Contracts;
 
 class SimilarDocumentsQuery
 {
-    private int|string $id;
+    /**
+     * @var int|string
+     */
+    private $id;
     private ?int $offset = null;
     private ?int $limit = null;
     private ?string $embedder = null;
@@ -15,13 +18,17 @@ class SimilarDocumentsQuery
     private ?bool $showRankingScoreDetails = null;
     private ?array $filter = null;
 
-    public function setId(string|int $id): SimilarDocumentsQuery
+    /**
+     * @param int|string $id
+     */
+    public function __construct($id)
     {
         $this->id = $id;
-
-        return $this;
     }
 
+    /**
+     * @param non-negative-int $offset
+     */
     public function setOffset(?int $offset): SimilarDocumentsQuery
     {
         $this->offset = $offset;
@@ -29,6 +36,9 @@ class SimilarDocumentsQuery
         return $this;
     }
 
+    /**
+     * @param positive-int $limit
+     */
     public function setLimit(?int $limit): SimilarDocumentsQuery
     {
         $this->limit = $limit;
@@ -36,6 +46,9 @@ class SimilarDocumentsQuery
         return $this;
     }
 
+    /**
+     * @param array<int, array<int, string>|string> $filter an array of arrays representing filter conditions
+     */
     public function setFilter(array $filter): SimilarDocumentsQuery
     {
         $this->filter = $filter;
@@ -43,6 +56,9 @@ class SimilarDocumentsQuery
         return $this;
     }
 
+    /**
+     * @param non-empty-string $embedder
+     */
     public function setEmbedder(string $embedder): SimilarDocumentsQuery
     {
         $this->embedder = $embedder;
@@ -50,6 +66,9 @@ class SimilarDocumentsQuery
         return $this;
     }
 
+    /**
+     * @param list<non-empty-string> $attributesToRetrieve an array of attribute names to retrieve
+     */
     public function setAttributesToRetrieve(array $attributesToRetrieve): SimilarDocumentsQuery
     {
         $this->attributesToRetrieve = $attributesToRetrieve;
@@ -57,6 +76,9 @@ class SimilarDocumentsQuery
         return $this;
     }
 
+    /**
+     * @param bool $showRankingScore boolean value to show ranking score
+     */
     public function setShowRankingScore(?bool $showRankingScore): SimilarDocumentsQuery
     {
         $this->showRankingScore = $showRankingScore;
@@ -64,6 +86,9 @@ class SimilarDocumentsQuery
         return $this;
     }
 
+    /**
+     * @param bool $showRankingScoreDetails boolean value to show ranking score details
+     */
     public function setShowRankingScoreDetails(?bool $showRankingScoreDetails): SimilarDocumentsQuery
     {
         $this->showRankingScoreDetails = $showRankingScoreDetails;
@@ -71,6 +96,9 @@ class SimilarDocumentsQuery
         return $this;
     }
 
+    /**
+     * @return array{id: int|string, offset: non-negative-int, limit: positive-int, filter: array<int, array<int, string>|string>, embedder: non-empty-string, attributesToRetrieve: list<non-empty-string>, showRankingScore: bool, showRankingScoreDetails: bool} SimilarDocumentsQuery converted to an array with non null fields
+     */
     public function toArray(): array
     {
         return array_filter([
@@ -82,6 +110,8 @@ class SimilarDocumentsQuery
             'attributesToRetrieve' => $this->attributesToRetrieve,
             'showRankingScore' => $this->showRankingScore,
             'showRankingScoreDetails' => $this->showRankingScoreDetails,
-        ], function ($item) { return null !== $item; });
+        ], function ($item) {
+            return null !== $item;
+        });
     }
 }
--- a/src/Contracts/SimilarDocumentsQuery.php
+++ b/src/Contracts/SimilarDocumentsQuery.php
@@ -87,7 +87,7 @@ class SimilarDocumentsQuery
     }
 
     /**
-     * @param bool $showRankingScoreDetails boolean value to show ranking score details
+     * @param bool|null $showRankingScoreDetails boolean value to show ranking score details
      */
     public function setShowRankingScoreDetails(?bool $showRankingScoreDetails): SimilarDocumentsQuery
     {
--- a/src/Contracts/SimilarDocumentsQuery.php
+++ b/src/Contracts/SimilarDocumentsQuery.php
@@ -77,7 +77,7 @@ class SimilarDocumentsQuery
     }
 
     /**
-     * @param bool $showRankingScore boolean value to show ranking score
+     * @param bool|null $showRankingScore boolean value to show ranking score
      */
     public function setShowRankingScore(?bool $showRankingScore): SimilarDocumentsQuery
     {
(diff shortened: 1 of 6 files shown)


src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Meilisearch\Contracts;
+
+class SimilarDocumentsQuery
+{
+    private int|string|null $id = null;
```
curquiza, line 9: Why `null`? this parameter is mandatory

tests/Endpoints/SimilarDocumentsTest.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        parent::setUp();
+
+        $this->index = $this->createEmptyIndex($this->safeIndexName());
+        $this->index->updateDocuments(self::DOCUMENTS);
+        $promise = $this->index->updateFilterableAttributes(['genre']);
+        $this->index->waitForTask($promise['taskUid']);
+    }
+
+    public function testBasicSearchWithFilters(): void
```
curquiza, line 25:
```suggestion
    public function testBasicSearchWithSimilarDocuments(): void
```

tests/Endpoints/SimilarDocumentsTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+        self::assertGreaterThanOrEqual(1, $response->getHitsCount());
+
+        $documentId = $response->getHit(0)['id'];
+        $response = $this->index->searchSimilarDocuments(
+            (new SimilarDocumentsQuery())
+                ->setId($documentId)
+        );
+
+        self::assertNotNull($response);
```
curquiza, line 37: Can we check we have some expected fields in the answer (specific to similar documents query) even if we don't check the value?

curquiza (changes requested):
Thank you very much for your PR @the-sinner  🙏 ❤️

curquiza:
@the-sinner do you think it would be relevant to create a SimilarDocumentsSearchResult file like https://github.com/meilisearch/meilisearch-php/blob/main/src/Search/SearchResult.php?

src/Contracts/SimilarDocumentsQuery.php
the-sinner: will fix this

tests/Endpoints/SimilarDocumentsTest.php
the-sinner: can you share the response format of this api, i could not find it.
i tried running the rc1 locally, however i was getting "Cannot find embedder with name `default`"

curquiza: I will ask my colleagues next week for having a basic tests with basic documents to make it test simply 😊

dureuill: Hello!

To use the similar endpoints, you **need** to:

1. enable the `vectorStore` experimental feature
2. declare at least one embedder in your configuration

For quick sanity tests, the simplest is probably to declare a `userProvided` embedder.

The settings could be:

```jsonc
"embedders": {
            "manual": {
                "source": "userProvided",
                "dimensions": 3,
            }
        }
```

Then declare a set of documents that contain the `_vectors.manual` field, for example:

```jsonc
[
        {
            "title": "Shazam!",
            "release_year": 2019,
            "id": "287947",
            // Three semantic properties:
            // 1. magic, anything that reminds you of magic
            // 2. authority, anything that inspires command
            // 3. horror, anything that inspires fear or dread
            "_vectors": { "manual": [0.8, 0.4, -0.5]},
        },
        {
            "title": "Captain Marvel",
            "release_year": 2019,
            "id": "299537",
            "_vectors": { "manual": [0.6, 0.8, -0.2] },
        },
        {
            "title": "Escape Room",
            "release_year": 2019,
            "id": "522681",
            "_vectors": { "manual": [0.1, 0.6, 0.8] },
        },
        {
            "title": "How to Train Your Dragon: The Hidden World",
            "release_year": 2019,
            "id": "166428",
            "_vectors": { "manual": [0.7, 0.7, -0.4] },
        },
        {
            "title": "All Quiet on the Western Front",
            "release_year": 1930,
            "id": "143",
            "_vectors": { "manual": [-0.5, 0.3, 0.85] },
        }
    ]
```

The example are from the [the engine test file](https://github.com/meilisearch/meilisearch/blob/main/meilisearch/tests/similar/mod.rs#L9) for the similar endpoints.

> can you share the response format of this api, i could not find it.

You're right I added the response format to [the public usage page](https://meilisearch.notion.site/Get-similar-documents-usage-540919ca755c4da0b7cdee273db3f290#f34dc66d893e42a7a6fe320efdc60c65), please feel free to ask for more information as needed :sunny:

curquiza:
@the-sinner can you update tests according to [dureuill message](https://github.com/meilisearch/meilisearch-php/pull/645#discussion_r1632879832)? 😇

the-sinner:
@curquiza i have added the test case shared by dureuill in the code.

curquiza (changes requested):
Thank you @the-sinner 

There is an issue with the tests

Capture d’écran 2024-06-11 à 10 05 06

the-sinner:
found the issue,
I am using Union types, but they are only supported after 8.0. [ref](https://www.php.net/releases/8.0/en.php#union-types)

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,87 @@
...
+    private int|string $id;
+    private ?int $offset = null;
+    private ?int $limit = null;
+    private ?string $embedder = null;
+    private ?array $attributesToRetrieve = null;
+    private ?bool $showRankingScore = null;
+    private ?bool $showRankingScoreDetails = null;
+    private ?array $filter = null;
+
+    public function setId(string|int $id): SimilarDocumentsQuery
```
norkunas, line 18: If ID is mandatory, why not require in constructor?

the-sinner:
@norkunas @curquiza 
For removing the union type, there are two ways:
1. make id string and only accept string values
2. remove type from it and manually check the type while assigning the value

which one do you think would be better ?

In the reference [link](https://meilisearch.notion.site/Get-similar-documents-usage-540919ca755c4da0b7cdee273db3f290) provided, the given example is using string ids, however the [linked](https://www.meilisearch.com/docs/learn/core_concepts/primary_key#formatting-the-document-id) documentation allows both string and integer keys.

norkunas:
Use phpdoc and that's all. People should respect the contract in the same way

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Meilisearch\Contracts;
+
+class SimilarDocumentsQuery
+{
+    private $id;
```
norkunas, line 9:
```suggestion
    /**
     * @var int|string
     */
    private $id;
```

src/Search/SimilarDocumentsSearchResult.php
```diff
@@ -0,0 +1,135 @@
...
+    private int $offset;
+    private int $limit;
+    private int $processingTimeMs;
+
+    private string $id;
+
+    public function __construct(array $body)
+    {
+        $this->id = $body['id'];
+        $this->hits = $body['hits'] ?? [];
```
norkunas, line 30: Is there are case when hits key would not exist? Because the line below you don't use null-coalescing like here

src/Search/SimilarDocumentsSearchResult.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public function getHitsCount(): int
+    {
+        return $this->hitsCount;
+    }
+
+    public function toArray(): array
+    {
+        $arr = [
```
norkunas, line 108: This variable is not necessary

the-sinner: fixed

src/Search/SimilarDocumentsSearchResult.php
the-sinner: fixed

src/Contracts/SimilarDocumentsQuery.php
the-sinner: fixed

src/Search/SimilarDocumentsSearchResult.php
```diff
@@ -0,0 +1,133 @@
...
+    {
+        return $this->id;
+    }
+
+    public function getHitsCount(): int
+    {
+        return $this->hitsCount;
+    }
+
+    public function toArray(): array
```
norkunas, line 106: I'd define return shape here

src/Search/SimilarDocumentsSearchResult.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->hits = $body['hits'];
+        $this->hitsCount = \count($body['hits']);
+        $this->processingTimeMs = $body['processingTimeMs'];
+        $this->offset = $body['offset'];
+        $this->limit = $body['limit'];
+        $this->estimatedTotalHits = $body['estimatedTotalHits'];
+    }
+
+    /**
+     * Return a new {@see SearchResult} instance.
```
norkunas, line 39: I don't see how this returns a new SearchResult, just modifies previous one :) it should clone itself then..
Also what's the point of this method, if you can directly call transformHits?

the-sinner: i will remove this method as the `transformHits` is public as well. I will add documentation on `transformHits` specifying what it is doing.

Also, is the `transformHits` needed or should the user write his own custom function to manipulate the search results ?

norkunas: I'd say not needed, unless we have a use case for this in the library itself.. butnas I understand you just copied from another class so maybe @curquiza should decide

src/Search/SimilarDocumentsSearchResult.php
```diff
@@ -0,0 +1,133 @@
...
+
+    public function transformHits(callable $callback): self
+    {
+        $this->hits = $callback($this->hits);
+        $this->hitsCount = \count($this->hits);
+
+        return $this;
+    }
+
+    public function getHit(int $key, $default = null)
```
norkunas, line 63:
```suggestion
    /**
     * @return array<string, mixed>|null
     */
    public function getHit(int $key): ?array
```

src/Search/SimilarDocumentsSearchResult.php
```diff
@@ -0,0 +1,133 @@
...
+            'hits' => $this->hits,
+            'hitsCount' => $this->hitsCount,
+            'processingTimeMs' => $this->processingTimeMs,
+            'offset' => $this->offset,
+            'limit' => $this->limit,
+            'estimatedTotalHits' => $this->estimatedTotalHits,
+        ];
+    }
+
+    public function toJSON(): string
```
norkunas, line 119: I think this method is also no needed

src/Search/SimilarDocumentsSearchResult.php
curquiza: If you think it's not needed, let's keep it simple, and remove it

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,102 @@
...
+
+    public function setEmbedder(string $embedder): SimilarDocumentsQuery
+    {
+        $this->embedder = $embedder;
+
+        return $this;
+    }
+
+    /**
+     * @param string[] $attributesToRetrieve an array of attribute names to retrieve
```
norkunas, line 61:
```suggestion
     * @param array<non-empty-string> $attributesToRetrieve an array of attribute names to retrieve
```

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,102 @@
...
+
+    public function setLimit(?int $limit): SimilarDocumentsQuery
+    {
+        $this->limit = $limit;
+
+        return $this;
+    }
+
+    /**
+     * @param string[][] $filter an array of arrays representing filter conditions
```
norkunas, line 44: Could this shape be more specific?

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * @param int|string $id
+     */
+    public function __construct($id)
+    {
+        $this->id = $id;
+    }
+
+    public function setOffset(?int $offset): SimilarDocumentsQuery
```
norkunas, line 29:
```suggestion
    /**
     * @param non-negative-int $offset
     */
    public function setOffset(?int $offset): SimilarDocumentsQuery
```

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        $this->offset = $offset;
+
+        return $this;
+    }
+
+    public function setLimit(?int $limit): SimilarDocumentsQuery
```
norkunas, line 36:
```suggestion
    /**
     * @param positive-int $limit
     */
    public function setLimit(?int $limit): SimilarDocumentsQuery
```

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,102 @@
...
+     * @param string[][] $filter an array of arrays representing filter conditions
+     */
+    public function setFilter(array $filter): SimilarDocumentsQuery
+    {
+        $this->filter = $filter;
+
+        return $this;
+    }
+
+    public function setEmbedder(string $embedder): SimilarDocumentsQuery
```
norkunas, line 53:
```suggestion
    /**
     * @param non-empty-string $embedder
     */
    public function setEmbedder(string $embedder): SimilarDocumentsQuery
```

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,102 @@
...
+
+    public function setShowRankingScoreDetails(?bool $showRankingScoreDetails): SimilarDocumentsQuery
+    {
+        $this->showRankingScoreDetails = $showRankingScoreDetails;
+
+        return $this;
+    }
+
+    /**
+     * @return array SimilarDocumentsQuery converted to an array with non null fields
```
norkunas, line 85: I still don't see shape defined :)

norkunas (changes requested):

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public function setAttributesToRetrieve(array $attributesToRetrieve): SimilarDocumentsQuery
+    {
+        $this->attributesToRetrieve = $attributesToRetrieve;
+
+        return $this;
+    }
+
+    /**
+     * @param bool $showRankingScore boolean value to show ranking score
```
norkunas, line 80:
```suggestion
     * @param bool|null $showRankingScore boolean value to show ranking score
```

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public function setShowRankingScore(?bool $showRankingScore): SimilarDocumentsQuery
+    {
+        $this->showRankingScore = $showRankingScore;
+
+        return $this;
+    }
+
+    /**
+     * @param bool $showRankingScoreDetails boolean value to show ranking score details
```
norkunas, line 90:
```suggestion
     * @param bool|null $showRankingScoreDetails boolean value to show ranking score details
```

norkunas (changes requested):

norkunas (approved):

curquiza:
Test failing are expected and due to the new RC

PR #656, opened 2024-07-23
Handle unexpected but valid JSON content-types

brunoocasali:
# Pull Request

## Related issue

When using Meilisearch analytics aka (meilisearch edge) users were facing weird decoding issues using this SDK. 

I found the reason was some particular routes on the edge were responding a valid body but they contained a content-type that was not recognized as valid by this SDK `application/json; charset=utf-8`.

## What does this PR do?

- Handle this use case
- Add a test to cover the new behavior


src/Http/Client.php
```diff
@@ -176,7 +176,7 @@ private function parseResponse(ResponseInterface $response)
             return null;
         }
 
-        if (!\in_array('application/json', $response->getHeader('content-type'), true)) {
+        if (!$this->isJSONResponse($response->getHeader('content-type'))) {
```
norkunas, line 179: Do we really need to check if it's json response? If we expect always to get json, then just delegate to json decoding, catch syntax error and just rethrow with meaningful message if needed?

brunoocasali: Yes, it should be always a JSON response, but I have no idea how we could do the way you're suggesting 😅

norkunas: Something like:

```diff
Index: src/Http/Client.php
===================================================================
diff --git a/src/Http/Client.php b/src/Http/Client.php
--- a/src/Http/Client.php	(revision a85bd4c4c54afc16f67c7bce66ccb789e810363e)
+++ b/src/Http/Client.php	(date 1721816875796)
@@ -176,16 +176,16 @@
             return null;
         }
 
-        if (!\in_array('application/json', $response->getHeader('content-type'), true)) {
-            throw new InvalidResponseBodyException($response, $response->getBody()->getContents());
-        }
-
         if ($response->getStatusCode() >= 300) {
             $body = $this->json->unserialize($response->getBody()->getContents()) ?? $response->getReasonPhrase();
 
             throw new ApiException($response, $body);
         }
 
-        return $this->json->unserialize($response->getBody()->getContents());
+        try {
+            return $this->json->unserialize($response->getBody()->getContents());
+        } catch (JsonDecodingException $e) {
+            throw new InvalidResponseBodyException($response, $response->getBody()->getContents());
+        }
     }
 }

```

brunoocasali: And I had a test that does not work as intended anymore:

```php
public function testInvalidResponseContentTypeThrowsException(): void
    {
        $httpClient = $this->createHttpClientMock(200, '<b>not json</b>', 'text/html');

        $client = new Client('https://localhost', null, $httpClient);

        $this->expectException(InvalidResponseBodyException::class);
        $this->expectExceptionMessage('Decoding payload to json');

        $client->get('/');
    }
```

In my mind the expectation before the exercise would catch that exception `InvalidResponseBodyException` because any other exception thrown before was swallowed by the `catch (Exception $e)`. So I didn't understand if this was only because the test was wrong or we would be changing the behavior if proceed with this change:

this is the error: 

`1) Tests\Http\ClientTest::testInvalidResponseContentTypeThrowsException
Failed asserting that exception of type "Meilisearch\Exceptions\JsonDecodingException" matches expected exception "Meilisearch\Exceptions\InvalidResponseBodyException". Message was: "Decoding payload to json failed: "Syntax error"." at`

brunoocasali: Maybe the exception has been raised before... Unfortunately, I don't have time to keep looking at this issue :|

brunoocasali: We could merge like this, then take a look at this later with more time, what do you think?

src/Http/Client.php
```diff
@@ -188,4 +188,20 @@ private function parseResponse(ResponseInterface $response)
...
+     *
+     * @param array $headerValues the array of header values to check
+     *
+     * @return bool true if any header value contains 'application/json', otherwise false
+     */
+    private function isJSONResponse(array $headerValues)
+    {
+        $filteredHeaders = array_filter($headerValues, function ($headerValue) {
+            return preg_match('/\bapplication\/json\b/', $headerValue);
+        });
```
norkunas, line 203:
```suggestion
        $filteredHeaders = array_filter($headerValues, function (string $headerValue) {
            return strpos($headerValue, 'application/json') !== false;
        });
```

norkunas (commented):
ok for merge, just no need for preg_match

Issue #659, opened 2024-07-29
[v1.10.0] Federated search


curquiza:
Related to https://github.com/meilisearch/integration-guides/issues/302

### Explanation of the feature

Pass new search parameters to `multiSearch` method to get one single list of result
Ensure the queries passed can accept the `federationOptions` search parameter
No need to handle the case of errors, Meilisearch does it

Usage: https://meilisearch.notion.site/v1-10-federated-search-698dfe36ab6b4668b044f735fb40f0b2?pvs=74

### TODO

- [x] Add a new search parameter named `federation` to the `multiSearch` method: an optional parameter to the `multiSearch` method
- [x] Add the capability to have a new search parameter to the search query named `federationOptions`
- [x] Add tests

⚠️ Make PRs pointing to `bump-meilisearch-v1.10.0` and NOT `main`. Please do 1 PR for all of these changes, and not several.

curquiza:
Closed by #663

Will be released once v1.10 of meilisearch is out

PR #662, opened 2024-07-30
Implements localized attributes

irevoire:
# Pull Request

> [!WARNING]  
> The tests won't work since the Meilisearch implementation is broken currently in rc0, see the following PR https://github.com/meilisearch/meilisearch/pull/4836
> I did the test on the branch that implemented the change though.

## Related issue
Fixes https://github.com/meilisearch/meilisearch-php/issues/660

## What does this PR do?
- [x] Add a new search parameter named `locales` to the `search` method
- [x] Ensure the `updateSettings` route can accept the new `localizedAttributes` setting
- [x] Create new methods `getLocalizedAttributes`, `updateLocalizedAttributes` and `resetLocalizedAttributes` corresponding to the sub settings `GET/PUT/DELETE` routes
- [x] Add tests

PR #663, opened 2024-07-31
Handle the new federation options

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch-php/issues/659

## What does this PR do?
- Create a new object, `federationOptions`, which can specify the weight of a query.
- Add a method to provide the federationOptions to a search request
- Adds the `federation` field parameter in the multi-search
- Add a test

Hey @norkunas, since you reviewed https://github.com/meilisearch/meilisearch-php/pull/662 and are definitely better than me in PHP I added you as a reviewer in this PR as well.
If you don't have the time don't worry we'll find another reviewer 👍

### DIFF
--- a/src/Contracts/SearchQuery.php
+++ b/src/Contracts/SearchQuery.php
@@ -282,7 +282,7 @@ class SearchQuery
             'showRankingScoreDetails' => $this->showRankingScoreDetails,
             'rankingScoreThreshold' => $this->rankingScoreThreshold,
             'distinct' => $this->distinct,
-            'federationOptions' => $this->federationOptions ? $this->federationOptions->toArray() : null,
+            'federationOptions' => $this->federationOptions !== null ? $this->federationOptions->toArray() : null,
         ], function ($item) { return null !== $item; });
     }
 }
--- a/src/Contracts/SearchQuery.php
+++ b/src/Contracts/SearchQuery.php
@@ -4,6 +4,25 @@ declare(strict_types=1);
 
 namespace Meilisearch\Contracts;
 
+class FederationOptions
+{
+    private ?float $weight = null;
+
+    public function setWeight(float $weight): FederationOptions
+    {
+        $this->weight = $weight;
+
+        return $this;
+    }
+
+    public function toArray(): array
+    {
+        return array_filter([
+            'weight' => $this->weight ?? null,
+        ], function ($item) { return null !== $item; });
+    }
+}
+
 class SearchQuery
 {
     private string $indexUid;
@@ -31,6 +50,7 @@ class SearchQuery
     private ?bool $showRankingScoreDetails = null;
     private ?float $rankingScoreThreshold = null;
     private ?string $distinct = null;
+    private ?FederationOptions $federationOptions = null;
 
     public function setQuery(string $q): SearchQuery
     {
@@ -199,6 +219,17 @@ class SearchQuery
     }
 
     /**
+     * This option is only available while doing a federated search.
+     * If used in another context an error will be returned by Meilisearch.
+     */
+    public function setFederationOptions(FederationOptions $federationOptions): SearchQuery
+    {
+        $this->federationOptions = $federationOptions;
+
+        return $this;
+    }
+
+    /**
      * This is an EXPERIMENTAL feature, which may break without a major version.
      * It's available from Meilisearch v1.3.
      * To enable it properly and use vector store capabilities it's required to activate it through the /experimental-features route.
@@ -251,6 +282,7 @@ class SearchQuery
             'showRankingScoreDetails' => $this->showRankingScoreDetails,
             'rankingScoreThreshold' => $this->rankingScoreThreshold,
             'distinct' => $this->distinct,
+            'federationOptions' => $this->federationOptions ? $this->federationOptions->toArray() : null,
         ], function ($item) { return null !== $item; });
     }
 }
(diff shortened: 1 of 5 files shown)


src/Contracts/SearchQuery.php
```diff
@@ -4,6 +4,25 @@
...
+        $this->weight = $weight;
+
+        return $this;
+    }
+
+    public function toArray(): array
+    {
+        return array_filter([
+            'weight' => $this->weight ?? null,
+        ], function ($item) { return null !== $item; });
```
norkunas, line 22:
```suggestion
        ], static function ($item) { return null !== $item; });
```

src/Contracts/SearchQuery.php
```diff
@@ -4,6 +4,25 @@
...
+    {
+        $this->weight = $weight;
+
+        return $this;
+    }
+
+    public function toArray(): array
+    {
+        return array_filter([
+            'weight' => $this->weight ?? null,
```
norkunas, line 21:
```suggestion
            'weight' => $this->weight,
```
weight is always defined or null, so no need for null-coalsecing

src/Endpoints/Delegates/HandlesMultiSearch.php
```diff
@@ -13,14 +13,16 @@ trait HandlesMultiSearch
     /**
      * @param list<\Meilisearch\Contracts\SearchQuery> $queries
      */
-    public function multiSearch(array $queries = [])
+    public function multiSearch(array $queries = [], array $multiSearchParams = [])
```
norkunas, line 16: would be nice to define `$multiSearchParams` shape

src/Endpoints/Delegates/HandlesMultiSearch.php
```diff
@@ -13,14 +13,16 @@ trait HandlesMultiSearch
...
         $body = [];
 
         foreach ($queries as $query) {
             $body[] = $query->toArray();
         }
 
-        return $this->http->post('/multi-search', ['queries' => $body]);
+        $payload = array_merge(['queries' => $body], $multiSearchParams);
+
+        return $this->http->post('/multi-search', $payload);
```
norkunas, line 26:
```suggestion
        return $this->http->post('/multi-search', ['queries' => $body, ...$multiSearchParams]);
```

irevoire: Oooh that’s nice!

src/Endpoints/Delegates/HandlesMultiSearch.php
curquiza: Hey @norkunas what do you mean by `$multiSearchParams shape`?

norkunas: hey. i mean what is possible to provide in it, so users would get autocomplete from IDE

norkunas: also `\Meilisearch\Contracts\SearchQuery` should be imported with use statement instead of using FQCN in `@param`

irevoire: > i mean what is possible to provide in it

By introducing a new object like I did with the `FederationOptions` or a comment would be enough?

norkunas: I looked at the notion and as I see federation options has to be included alongside the queries, so not sure why FederationOptions was added to SearchQuery?

irevoire: The `federation` field is included alongside the queries, but then the `FederationOptions` is included in each search query and currently only contains a `weight` to make one query more important than another.
But in the future, they should contain more parameters.

norkunas: If they can differ then I suggest to add `MultiSearchFederation` with `limit`/`offset`/`reranking` and other available options that I don't know and document the `$multiSearchParams` as: `array{federation: MultiSearchFederation}`

irevoire: Seems good to me, will work on it later

norkunas (changes requested):

src/Contracts/SearchQuery.php
```diff
@@ -4,6 +4,53 @@
 
 namespace Meilisearch\Contracts;
 
+class MultiSearchFederation
```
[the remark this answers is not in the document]
irevoire: Can I put the two Federation related class in the same file or should I make one file for each?

src/Endpoints/Delegates/HandlesMultiSearch.php
```diff
@@ -13,14 +14,19 @@ trait HandlesMultiSearch
     /**
      * @param list<\Meilisearch\Contracts\SearchQuery> $queries
```
norkunas, line 15:
```suggestion
     * @param list<SearchQuery> $queries
```

src/Endpoints/Delegates/HandlesMultiSearch.php
```diff
@@ -5,6 +5,7 @@
 namespace Meilisearch\Endpoints\Delegates;
 
 use Meilisearch\Contracts\Http;
+use Meilisearch\Contracts\MultiSearchFederation;
```
norkunas, line 8:
```suggestion
use Meilisearch\Contracts\MultiSearchFederation;
use Meilisearch\Contracts\SearchQuery;
```

norkunas, line 8:
```suggestion
use Meilisearch\Contracts\MultiSearchFederation;
use Meilisearch\Contracts\SearchQuery;
```

norkunas (approved):

PR #678, opened 2024-09-11
Cast body to a string to force rewind of stream

grizzm0:
# Pull Request

## What does this PR do?
- Rewinds the PSR-7 StreamInterface

When reading a PSR-7 StreamInterface with getContents() the stream is not automatically rewound. If you for example attach a middleware in Guzzle that reads the StreamInterface but does not rewind it afterwards any calls afterwards will read from the end of the stream resulting in an empty string.

However, then casting the StreamInterface to a string using __toString() then the stream MUST attempt to seek to the beginning of the stream. See https://www.php-fig.org/psr/psr-7/#34-psrhttpmessagestreaminterface

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -177,16 +177,16 @@ class Client implements Http
         }
 
         if (!$this->isJSONResponse($response->getHeader('content-type'))) {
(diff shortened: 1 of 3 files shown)


norkunas (approved):

brunoocasali:
bors merge

PR #686, opened 2024-10-04
Cast federation payload explicilty to an object

### LINKED ISSUE #685: Federated multi-search example does not work
**Description**
I'm using the federated multi-search feature of Meilisearch. I use the code example from the docs:

```php
$client->multiSearch([
    (new SearchQuery())
      ->setIndexUid('movies'))
      ->setQuery('batman'),
    (new SearchQuery())
      ->setIndexUid('comics')
      ->setQuery('batman'),
  ],
  (new MultiSearchFederation())
);
```
However, this results in the following error:

afbeelding

This error **only** appears when I pass an empty MultiSearchFederation() instance, like in the docs. The error **does not** appear when I set a property of MultiSearchFederation, like `(new MultiSearchFederation())->setLimit(limit: 100)`.

**Possible solution**
See my PR

thijskuilman:
# Pull Request

## Related issue
Fixes #685

## What does this PR do?
- It casts the `federation` payload explicitly to an object. This solves the error that appears when no `MultiSearchFederation` properties have been set. 

## PR checklist

### DIFF
--- a/src/Endpoints/Delegates/HandlesMultiSearch.php
+++ b/src/Endpoints/Delegates/HandlesMultiSearch.php
@@ -25,7 +25,7 @@ trait HandlesMultiSearch
 
         $payload = ['queries' => $body];
         if (null !== $federation) {
-            $payload['federation'] = $federation->toArray();
+            $payload['federation'] = (object) $federation->toArray();
         }
 
         return $this->http->post('/multi-search', $payload);


norkunas:
Please add test then to avoid regressions in future

curquiza (changes requested):
@thijskuilman 

thank you for your PR ❤️

I agree with norkunas, can you add a test with an empty MultiSearchFederation to ensure it works all the time?

thijskuilman:
@norkunas @curquiza I have added a test!

norkunas:
@thijskuilman not sure that we need a functional test here :) better would be to assert on that what was really passed to http client, because now a test without assertions doesn't prove anything

norkunas:
IMHO we should a dev requirement for symfony/http-client, which allows easier testing than the guzzle.

Then your test could look like this:

```
$mockHttpClient = new \Symfony\Component\HttpClient\MockHttpClient(static function (string $method, string $url, array $options) {
    // assert on method, url, options['body'] or whatever what was passed

    return new MockResponse();
});
$httpClient = new \Symfony\Component\HttpClient\Psr18Client($mockHttpClient);

$meilisearch = new \Meilisearch\Client('http://meilisearch', 'api_key', $httpClient);
$meilisearch->multiSearch([], new MultiSearchFederation());
```

code is just an idea and is not validated :)

norkunas:
My point was that unit test will be faster than the functional one :)

src/Endpoints/Delegates/HandlesMultiSearch.php
```diff
@@ -25,7 +25,7 @@ public function multiSearch(array $queries = [], ?MultiSearchFederation $federat
 
         $payload = ['queries' => $body];
         if (null !== $federation) {
-            $payload['federation'] = $federation->toArray();
+            $payload['federation'] = (object) $federation->toArray();
```
ruhulfbr, line 28: Why is typecasting necessary? You could use `$payload['federation'] = $federation;` and it would work just fine.

ruhulfbr:
@curquiza  Could you please review and process this PR or provide a solution? This is an important update.

curquiza:
Hello @ruhulfbr 
we have a lot on our plate now, but @Strift will review your PR when he will have more time for this
In the meantime if you have time @norkunas, you can review and merge of course. If you cannot, no worries, we will take care of it later

norkunas:
I've already reviewed and my point still stands - it'd enough for unit test

norkunas:
Closing as it's already been fixed

PR #688, opened 2024-10-08
Multi search facet distribution

### LINKED ISSUE #681: [v1.11.0] Support facet distribution for federated search
Related to https://github.com/meilisearch/integration-guides/issues/303

* Request facet distributions and facet stats in your federated search, by using `federation.facetsByIndex` in the `POST /multi-search` route.
* Merge the returned facets in a single facet distribution and stats that is global to the entire request, by using `federation.mergeFacets` in the `POST /multi-search` route.

Usage: https://meilisearch.notion.site/v1-11-Federated-search-59b30e03383c40729d7541a3dffb0069

### TODO

- [ ] Ensure the method `multiSearch` can accept the `facetsByIndex` option
- [ ] Ensure the method `multiSearch` can accept the `mergeFacets` option
- [ ] Add tests

⚠️ Make PRs pointing to `bump-meilisearch-v1.11.0` and NOT `main`. Please do 1 PR for all of these changes, and not several.

ManyTheFish:
Fixes #681

### DIFF
--- a/src/Contracts/MultiSearchFederation.php
+++ b/src/Contracts/MultiSearchFederation.php
@@ -21,6 +21,9 @@ class MultiSearchFederation
      */
     private ?array $facetsByIndex = null;
 
+    /**
+     * @var array{maxValuesPerFacet: positive-int}|null
+     */
     private ?array $mergeFacets = null;
 
     /**
@@ -60,6 +63,8 @@ class MultiSearchFederation
     }
 
     /**
+     * @param array{maxValuesPerFacet: positive-int} $mergeFacets
+     *
      * @return $this
      */
     public function setMergeFacets(array $mergeFacets): self
@@ -74,7 +79,7 @@ class MultiSearchFederation
      *     limit?: non-negative-int,
      *     offset?: non-negative-int,
      *     facetsByIndex?: array<non-empty-string, list<non-empty-string>>,
-     *     mergeFacets?: array,
+     *     mergeFacets?: array{maxValuesPerFacet: positive-int},
      * }
      */
     public function toArray(): array
--- a/src/Contracts/MultiSearchFederation.php
+++ b/src/Contracts/MultiSearchFederation.php
@@ -17,6 +17,13 @@ class MultiSearchFederation
     private ?int $offset = null;
 
     /**
+     * @var array<non-empty-string, list<non-empty-string>>|null
+     */
+    private ?array $facetsByIndex = null;
+
+    private ?array $mergeFacets = null;
+
+    /**
      * @param non-negative-int $limit
      *
      * @return $this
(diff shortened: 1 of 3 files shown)


src/Contracts/MultiSearchFederation.php
```diff
@@ -16,6 +16,13 @@ class MultiSearchFederation
      */
     private ?int $offset = null;
 
+    /**
+     * @var array<non-empty-string, list<non-empty-string>>|null
+     */
+    private ?array $facetsByIndex = null;
+
+    private ?array $mergeFacets = null;
```
norkunas, line 24: What's the array shape?

ManyTheFish: string -> any value

norkunas: list or hash? then it should be documented as `@var list<string>` or `@var array<string>` :)

ManyTheFish:
```suggestion
    /**
     * @var array<non-empty-string>|null
     */
    private ?array $mergeFacets = null;
```

something like this?

norkunas: yes :) i found this:

image

so it can be also an integer, right?

ManyTheFish: mergeFacets is an object with only one field so far, but we should be able to support any kind of value depending on the evolution of this setting

norkunas: just thinking out loud: should `mergeFacets` be documented as `array{maxValuesPerFacet: int}` as looking at the implementation PR there is only this single option

norkunas: yes, gh didn't show your comment before posting mine :) we can evolve php docs after new features are merged.
with this shape definition user will get IDE autocompletion :)

ManyTheFish: and if the settings ios changed, we would just change this into `array{maxValuesPerFacet: int, newSetting: string}`?

norkunas: we could even make a new DTO for `mergeFacets`

norkunas: > and if the settings ios changed, we would just change this into:
> 
> ```
> array{maxValuesPerFacet: int, newSetting: string}
> ```
> 
> ?

yes

norkunas: but I'd prefer a new DTO :)

ManyTheFish: Ok so to summarize, for now we add the following code:
```suggestion
    /**
     * @var array{maxValuesPerFacet: int}|null
     */
    private ?array $mergeFacets = null;
```

and if, in the future, the `mergeFacets` becomes more complex, we will create a DTO.

norkunas: Ok

norkunas: just a nitpick, but maybe `@var array{maxValuesPerFacet: positive-int}|null`

norkunas (approved):

ManyTheFish (commented):
Thank you @norkunas for the review 👍 , merging this PR

PR #698, opened 2024-12-04
Add parameter to get tasks in reverse order

### LINKED ISSUE #693: [v1.12.0] `reverse` parameter when querying tasks
Related to https://github.com/meilisearch/integration-guides/issues/307

Add a new parameter in `GET /tasks` route: `reverse`

Usage: https://meilisearch.notion.site/Usage-Reverse-the-order-of-the-task-queue-1364b06b651f81dcb618e8409d30429b

### TODO

- [ ] Add a new query parameter associated to the `getTasks` method
- [ ] Add tests

⚠️ Make PRs pointing to `bump-meilisearch-v1.12.0` and NOT `main`. Please do 1 PR for all of these changes, and not several.

Strift:
# Pull Request

## Related issue
Fixes #693

### DIFF
--- a/src/Contracts/TasksQuery.php
+++ b/src/Contracts/TasksQuery.php
@@ -23,6 +23,11 @@ class TasksQuery
     private ?array $canceledBy = null;
 
     /**
+     * @var bool|null
+     */
+    private ?bool $reverse = null;
+
+    /**
      * @return $this
      */
     public function setFrom(int $from): self
@@ -54,6 +59,16 @@ class TasksQuery
         return $this;
     }
 
+    /**
+     * @return $this
+     */
+    public function setReverse(bool $reverse): self
+    {
+        $this->reverse = $reverse;
+
+        return $this;
+    }
+
     public function toArray(): array
     {
         return array_filter(
@@ -63,6 +78,7 @@ class TasksQuery
                     'from' => $this->from,
                     'limit' => $this->limit,
                     'canceledBy' => $this->formatArray($this->canceledBy),
+                    'reverse' => $this->reverse,
                 ]
             ), static function ($item) { return null !== $item; }
         );
--- a/src/Contracts/TasksQuery.php
+++ b/src/Contracts/TasksQuery.php
@@ -22,9 +22,6 @@ class TasksQuery
      */
     private ?array $canceledBy = null;
 
-    /**
-     * @var bool|null
-     */
     private ?bool $reverse = null;
 
     /**
@@ -78,7 +75,7 @@ class TasksQuery
                     'from' => $this->from,
                     'limit' => $this->limit,
                     'canceledBy' => $this->formatArray($this->canceledBy),
-                    'reverse' => $this->reverse !== null ? json_encode($this->reverse) : null,
+                    'reverse' => null !== $this->reverse ? json_encode($this->reverse) : null,
                 ]
             ), static function ($item) { return null !== $item; }
         );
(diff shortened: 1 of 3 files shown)


src/Contracts/TasksQuery.php
```diff
@@ -63,6 +75,7 @@ public function toArray(): array
                     'from' => $this->from,
                     'limit' => $this->limit,
                     'canceledBy' => $this->formatArray($this->canceledBy),
+                    'reverse' => null !== $this->reverse ? json_encode($this->reverse) : null,
```
norkunas, line 78: https://github.com/meilisearch/meilisearch-php/blob/ef1769975bcc3b97057dc6ee3ed3f733a8ec7f0e/src/Contracts/DocumentsQuery.php#L119

avoid json_encode to be consistent :)

tests/Contracts/TasksQueryTest.php
```diff
@@ -109,4 +109,11 @@ public function testSetFrom(): void
 
         self::assertSame(['from' => 1], $data->toArray());
     }
+
+    public function testSetReverse(): void
+    {
+        $data = (new TasksQuery())->setReverse(true);
+
+        self::assertSame(['reverse' => json_encode(true)], $data->toArray());
```
norkunas, line 117:
```suggestion
        self::assertSame(['reverse' => 'true'], $data->toArray());
```

tests/Endpoints/TasksTest.php
```diff
@@ -140,6 +140,19 @@ public function testCancelTasksWithFilter(): void
         self::assertSame('succeeded', $response['status']);
     }
 
+    public function testGetAllTasksInReverseOrder(): void
+    {
+        $date = new \DateTime();
```
norkunas, line 145:
```suggestion
        $date = new \DateTimeImmutable();
```

tests/Contracts/TasksQueryTest.php
```diff
@@ -109,4 +109,11 @@ public function testSetFrom(): void
...
+    }
```
norkunas, line 118:
```suggestion
    /**
     * @testWith [true, "true"]
     *                     [false, "false"]
     */
    public function testSetReverse(bool $reverse, string $expected): void
    {
        $data = (new TasksQuery())->setReverse($reverse);

        self::assertSame(['reverse' => $expected], $data->toArray());
    }
```

src/Contracts/TasksQuery.php
Strift: Thank you for the suggestion!

Strift:
Hello @norkunas and thanks for the early review 👋  I forgot to set the PR as a draft.

I applied all your suggestions. Thanks! We're now ready for a new review. 

It looks like the test is flaky though, because it failed at first; and worked when I re-ran it. I'm not sure how I can make it better. If you have feedback, it's welcome.

norkunas:
> Hello @norkunas and thanks for the early review 👋 I forgot to set the PR as a draft.

Hey :wave: 

> It looks like the test is flaky though, because it failed at first; and worked when I re-ran it. I'm not sure how I can make it better.

Which tasks does it retrieve? Because I don't see that index would be seeded?

norkunas:
Oh looks like that after each test the index is not cleared, so it sees previous tests data :expressionless:

norkunas (approved):
> Would you recommend updating the tests or do you think it's sufficient as is?

I'd at least add a (todo) note that these tests should be refactored to cleanup after themself

PR #699, opened 2024-12-04
Add batches endpoints

### LINKED ISSUE #695: [v1.12.0] New batch routes
Related to https://github.com/meilisearch/integration-guides/issues/307

- New `GET /batches` & `GET /batches/:uid` routes
- New field in `task` object: `batchUid`

Usage: https://meilisearch.notion.site/Enhance-visibility-on-batched-tasks-1194b06b651f810b8fe0fab5d72846a8?pvs=74

### TODO

👉 Check carefully the usage above to ensure your new methods and objects integrate all the functionnalities

- [ ] Create two new methods `getBatch()` and `getBatches()` triggering the appropriate new route (check out `getTask` and `getTasks` since the behavior will be pretty similar)
- [ ] A new `batch` object will probably be needed to be created.
- [ ] Ensure the `task` object can accept the `batchUid` field
- [ ] Add tests

⚠️ Make PRs pointing to `bump-meilisearch-v1.12.0` and NOT `main`. Please do 1 PR for all of these changes, and not several.

Strift:
# Pull Request

## Related issue
Fixes #695

### DIFF
--- a/src/Contracts/BatchesResults.php
+++ b/src/Contracts/BatchesResults.php
@@ -6,89 +6,89 @@ namespace Meilisearch\Contracts;
 
 class BatchesResults extends Data
 {
-  /**
-   * @var non-negative-int
-   */
-  private int $next;
+    /**
+     * @var non-negative-int
+     */
+    private int $next;
 
-  /**
-   * @var non-negative-int
-   */
-  private int $limit;
+    /**
+     * @var non-negative-int
+     */
+    private int $limit;
 
-  /**
-   * @var non-negative-int
-   */
-  private int $from;
+    /**
+     * @var non-negative-int
+     */
+    private int $from;
 
-  /**
-   * @var non-negative-int
-   */
-  private int $total;
+    /**
+     * @var non-negative-int
+     */
+    private int $total;
 
-  public function __construct(array $params)
-  {
-    parent::__construct($params['results'] ?? []);
+    public function __construct(array $params)
+    {
+        parent::__construct($params['results'] ?? []);
 
-    $this->from = $params['from'] ?? 0;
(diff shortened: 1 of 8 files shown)


src/Contracts/BatchesResults.php
```diff
@@ -0,0 +1,94 @@
...
+    private int $from;
+
+    /**
+     * @var non-negative-int
+     */
+    private int $total;
+
+    public function __construct(array $params)
+    {
+        parent::__construct($params['results'] ?? []);
```
norkunas, line 31: Are there any cases when results will not be set?

src/Endpoints/Delegates/HandlesBatches.php
```diff
@@ -0,0 +1,28 @@
...
+    protected Batches $batches;
+
+    public function getBatch($uid): array
+    {
+        return $this->batches->get($uid);
+    }
+
+    public function getBatches(?BatchesQuery $options = null): BatchesResults
+    {
+        $query = isset($options) ? $options->toArray() : [];
```
norkunas, line 22:
```suggestion
        $query = null !== $options ? $options->toArray() : [];
```

src/Contracts/BatchesResults.php
Strift: As far as I understand, the API should always return a `results` array. 

But I saw in the `TasksResults` class that there was a check for the field being missing. So I used the same logic.

norkunas: So it's a dead code then :slightly_smiling_face:

Strift: I updated this line ✅

norkunas (approved):

PR #701, opened 2024-12-05
Make `hitsCount` always the number of `hits` (and not `totalHits` value)

### LINKED ISSUE #700: hitsCount is the same as totalHits if estimatedTotalHits is not set
When the response from a search request doesn't include the `estimatedTotalHits` parameter, the `SearchResults::hitsCount` value is set to `totalHits`.

```
{
  "hits": [...],
  "hitsCount": 354, <---- Should be 4 in this case
  "semanticHitCount": 0,
  "hitsPerPage": 10,
  "page": 36,
  "totalPages": 36,
  "totalHits": 354,
  "processingTimeMs": 0,
  "numberedPagination": true,
  "query": "test",
  "facetDistribution": [],
  "facetStats": [],
  "raw": {
    "hits": [...],
    "query": "test",
    "processingTimeMs": 0,
    "hitsPerPage": 10,
    "page": 36,
    "totalPages": 36,
    "totalHits": 354
  }
}
```

I think that `hitsCount` should _always_ be the count of hits for the current page instead.

johnnynotsolucky:
# Pull Request

## Related issue
Fixes #700 

## What does this PR do?
- Ensures that `SearchResults::hitsCount` is consistent regardless of whether `estimatedTotalHits` is set.

## PR checklist

### DIFF
--- a/src/Search/SearchResult.php
+++ b/src/Search/SearchResult.php
@@ -58,7 +58,6 @@ class SearchResult implements \Countable, \IteratorAggregate
             $this->offset = $body['offset'];
             $this->limit = $body['limit'];
             $this->estimatedTotalHits = $body['estimatedTotalHits'];
-            $this->hitsCount = \count($body['hits']);
         } else {
             $this->numberedPagination = true;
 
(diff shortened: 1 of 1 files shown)


curquiza (commented):
Hello

This is more a maintenance change here, nothing that impacts the product because `totalHits` is already supposed to always be the count of `hits`

Did you ask for this change because you noticed `totalHits` was different from the number of `hits` @johnnynotsolucky?

johnnynotsolucky:
Hi @curquiza, I noticed that when `estimatedTotalHits` is _not_ set, then `hitsCount` has the same value as `totalHits`, which I think isn't correct. I would expect the value of `hitsCount` to always be the count of hits on the current page of results, whereas `totalHit` is the count or estimated count of hits across all pages.

curquiza (commented):
Sorry, I did not understand immediately the problem,

it's indeed a bug here!

`hitsCount` is the number of `hits`, and `totalHits` the total exhaustive number of `hits`

curquiza (approved):
thank you for this bug fix

bors merge

PR #721, opened 2025-01-31
Allow instantiating the Meilisearch client with custom HTTP headers

### LINKED ISSUE #720: Add option to pass headers directly in the Client builder - For example to easily pass X-MS-USER-ID
**Description**
I need to pass X-MS-USER-ID to track my user because I use search in my backend and not my front end. No easy way to pass them at Meilisearch Client build. And I do not want to pass a Guzzle client at build time.

wandyou:
# Pull Request

## Related issue
Fixes #720 

## What does this PR do?
- Fixes the issue by introducing a new option in the meilisearch client builder.
- Passes it down to the http client
- Merges the headers passed in the option with the other headers already configured (ex: bearer)

## Improvement
- This solution is not the cleanest because if forces us to null all the optional parameters before it (if we don't use them).
 For example : `new Meilisearch(MEILISEARCH_HOST, MEILISEARCH_SEARCH_KEY, null, null, [], null, ["X-MS-USER-ID" => example])` => But it was the fastest solution and the less destructive.
- An other way would be to create a method in the Meilisearch Client object `addHeader()`
- An other way would be to pass all Meilisearch Client options in the form of an array, allowing us to only define used values in any order we like.

## PR checklist
- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?

## Extra info
First contribution of mine to this repo, maybe I missed some subtilities.

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying custom HTTP headers when creating a client instance.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -47,9 +47,10 @@ class Client
         ?ClientInterface $httpClient = null,
         ?RequestFactoryInterface $requestFactory = null,
         array $clientAgents = [],
-        ?StreamFactoryInterface $streamFactory = null
+        ?StreamFactoryInterface $streamFactory = null,
+        $headers = []
     ) {
-        $this->http = new MeilisearchClientAdapter($url, $apiKey, $httpClient, $requestFactory, $clientAgents, $streamFactory);
+        $this->http = new MeilisearchClientAdapter($url, $apiKey, $httpClient, $requestFactory, $clientAgents, $streamFactory, $headers);
         $this->index = new Indexes($this->http);
         $this->health = new Health($this->http);
         $this->version = new Version($this->http);
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -41,7 +41,8 @@ class Client implements Http
         ?ClientInterface $httpClient = null,
         ?RequestFactoryInterface $reqFactory = null,
         array $clientAgents = [],
-        ?StreamFactoryInterface $streamFactory = null
+        ?StreamFactoryInterface $streamFactory = null,
+        $headers = []
     ) {
         $this->baseUrl = $url;
         $this->http = $httpClient ?? Psr18ClientDiscovery::find();
@@ -53,6 +54,7 @@ class Client implements Http
         if (null !== $apiKey && '' !== $apiKey) {
             $this->headers['Authorization'] = \sprintf('Bearer %s', $apiKey);
         }
+        $this->headers = array_merge($this->headers, $headers);
         $this->json = new Json();
     }
 


norkunas:
You can pass a http client configured with your headers, as it's a http client responsibility 🤔

wandyou:
I get it, and that's how I am solving this issue right now. 

But the `X-MS-USER-ID` header is specific to meilisearch's analytics and I find it more appealing and "light" to just be able to pass my unique user identifier while calling `new MeilisearchClient()`, then passing an instance of a full http client, add configuration options, to "just" give an Analytics header.

In fact in some cases, I would like to be able to change the user-id passed in the header on the go. Maybe with a `$client->setHeader()`or `$client->setUserId()`. Because I want to perform a search while being the user and an other while being the server. And having to instanciate two meilisearch clients seems overkill.

It is obvious to me that few only projects will need to pass a "custom" http client to the MeilisearchClient, but all of them would like to take advantage of precise analytics. And therefore pass a "X-MS-USER-ID". So I find it interesting to make that task more "accessible".

Maybe the right thing to do is not to touch the headers upfront, but to expose a method like `setAnalyticsUserId()`and under the hood it sets the header.

I know analytics is in beta for now, so all of this is maybe new and not much used. But as a developer, not needing to customise my http client. I would prefer having a simple way to pass this info. I did not find any Analytics methods in the php doc neither. For example to report a clic on a search hit, or a conversion.

Maybe I am missing a key concept here and not seeing the big picture, and maybe I am not expressing myself clearly enough because english is not my native language. So obviously feel free to jump in and explain why it would not be the way to go.

Thanks

wandyou:
I may have not been clear enough on the fact that I want to simplify the implementation of `X-MS-USER-ID` because in my opinion, analytics is a mandatory option for most use cases. And most devs will want to pass it, so why not make it dev friendly ?

Obviously if I want to configure and use custom headers, the best practice would be to pass a custom http client.

I get that adding an optional headers array, whilst seeming the easiest implementation on my end, is not the best suited for the long run.

Maybe a method like `setAnalyticsUserId()`on the client would ? Or any other suggestions ?

norkunas:
> And having to instanciate two meilisearch clients seems overkill.

I wouldn't say so. Stateful singleton is more harmful than instantiating single per user :)

I see in docs it is explicitly related to search endpoint. So I think this could be accepted as a SearchQuery option, but not as a global setting for whole Meilisearch client

wandyou:
> I see in docs it is explicitly related to search endpoint. So I think this could be accepted as a SearchQuery option, but not as a global setting for whole Meilisearch client

Well as of for now, analytics is only done on searches, so yes indeed it would make more sense to add it as a Search Query option.

And regarding other analytics endpoints ? Like hit clicking, conversions, etc ? Would be nice to implement those as well.

norkunas:
> And regarding other analytics endpoints ? Like hit clicking, conversions, etc ? Would be nice to implement those as well.

Yes, for this you could open another issue :)

Strift (dismissed):
Sorry for the delayed review. I would improve types if possible, but this can be handled in another PR if we want to merge ASAP.

edit: sorry I approved too quick, adding tests would be welcome here

Strift (changes requested):
I recommend adding tests to verify that this works as intended. 

I would also improve types as recommended by Copilot

PR #727, opened 2025-02-23
add mixed return type

### LINKED ISSUE #726: deprecation notice
User Deprecated: Method "ArrayAccess::offsetGet()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Meilisearch\Contracts\Data" now to avoid errors or add an explicit @return annotation to suppress this message.

I'll submit a PR to add mixed.

tacman:
# Pull Request

## Related issue
Fixes #726 

## What does this PR do?

Gets rid of the deprecation notice.  But I forget when 'mixed' was introduced and if this library is supporting a version of php that doesn't allow this.

### DIFF
--- a/src/Contracts/Data.php
+++ b/src/Contracts/Data.php
@@ -29,7 +29,7 @@ class Data implements \ArrayAccess, \Countable, \IteratorAggregate
     }
 
     #[\ReturnTypeWillChange]
-    public function offsetGet($offset)
+    public function offsetGet($offset): mixed
     {
         if (isset($this->data[$offset])) {
             return $this->data[$offset];


norkunas:
Would it work to resolve deprecation adding only phpdoc return typehint?

tacman:
Probably.  I see that 'mixed' is PHP 8 only.

PHP 7.4 reached EOL in 2022.  What about simply bumping a new release and taking advantage of all the PHP 8 goodness?

Strift:
We can pick this up after merging #735 which drops PHP versions < 8.

norkunas:
This is already in main branch, closing.

PR #730, opened 2025-03-19
add NDJSON Serialization class

TheDigitalOrchard:
# Pull Request

## What does this PR do?
- adds NDJSON class for serialization

## PR checklist
- [Y ] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [ Y] Have you read the contributing guidelines?
- [ Y] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -13,6 +13,7 @@ use Meilisearch\Exceptions\InvalidResponseBodyException;
 use Meilisearch\Exceptions\JsonDecodingException;
 use Meilisearch\Exceptions\JsonEncodingException;
 use Meilisearch\Http\Serialize\Json;
+use Meilisearch\Http\Serialize\Ndjson;
 use Meilisearch\Meilisearch;
 use Psr\Http\Client\ClientExceptionInterface;
 use Psr\Http\Client\ClientInterface;
@@ -30,7 +31,7 @@ class Client implements Http
     /** @var array<string,string> */
     private array $headers;
     private string $baseUrl;
-    private Json $json;
+    private Json|Ndjson $json;
 
     /**
      * @param array<int, string> $clientAgents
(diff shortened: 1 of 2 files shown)


src/Http/Client.php
```diff
@@ -30,7 +31,7 @@ class Client implements Http
     /** @var array<string,string> */
     private array $headers;
     private string $baseUrl;
-    private Json $json;
+    private Json|Ndjson $json;
```
norkunas, line 34: this should be `Meilisearch\Http\Serialize\SerializerInterface` instead.

src/Http/Client.php
```diff
@@ -83,9 +91,12 @@ public function post(string $path, $body = null, array $query = [], ?string $con
     {
         if (!\is_null($contentType)) {
             $this->headers['Content-type'] = $contentType;
+        } elseif (str_ends_with($body, "}\n")) {
+            $this->headers['Content-type'] = 'application/x-ndjson';
+            $body = $this->json(new Ndjson)->serialize($body);
```
norkunas, line 96: I don't this is a good idea to switch format on the fly. Serializer should be preconfigured imho, or if contentType is manually set, then just instantiate other serializer without overriding the client state

src/Http/Serialize/Ndjson.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Meilisearch\Http\Serialize;
+
+use Meilisearch\Exceptions\JsonDecodingException;
+use Meilisearch\Exceptions\JsonEncodingException;
+
+class Ndjson implements SerializerInterface
```
norkunas, line 10:
```suggestion
final class Ndjson implements SerializerInterface
```

src/Http/Client.php
```diff
@@ -181,25 +201,25 @@ private function parseResponse(ResponseInterface $response)
...
      * @param array $headerValues the array of header values to check
      *
-     * @return bool true if any header value contains 'application/json', otherwise false
+     * @return bool true if any header value contains 'application/json' or 'application/x-ndjson', otherwise false
      */
     private function isJSONResponse(array $headerValues): bool
     {
         $filteredHeaders = array_filter($headerValues, static function (string $headerValue) {
-            return false !== strpos($headerValue, 'application/json');
+            return str_contains($headerValue, 'application/json') || str_contains($headerValue, 'application/x-ndjson');
```
norkunas, line 222: str_contains is available only from php 8, this library supports php 7.4, so you cannot use it yet

norkunas (changes requested):

PR #735, opened 2025-04-09
Task object

### LINKED ISSUE #703: Replace array return types from api to object
**Description**
Currently all enqueued tasks are returned as an array without defined structure, so basically we cannot be sure that is safe to access any data from it.
As I've seen structure can differ per task status, but I could be mistaken.
This idea came up to my mind after I've seen https://github.com/meilisearch/meilisearch-js/blob/main/src/enqueued-task.ts

**Basic example**
```php
class EnqueuedTask implements \ArrayAccess
{
  private int $taskUid;
  private string indexUid;
  private string $status; //  'succeeded'|'processing'|'failed'|'enqueued'|'canceled'; // this could be class constants and later migrated to enums after bumping up php requirement to 8+
  private string $type; // ...
  private DateTimeImmutable $enqueuedAt;
}
```
This could act as a base, and if response data is really different per task status, then we could extend this object into many ones with more additional information. another approach of course would be just add all data to `EnqueuedTask` and for consumers to check status before accessing it

**Other**
This is related to #573

To break things less the returned lets say `EnqueuedTask` object could implement the `ArrayAccess` interface so array access would still work, it would just break things if someone has extended any of Endpoint classes.

norkunas:
# Pull Request

## Related issue
Fixes #703

## What does this PR do?
- Bumps PHP requirement to 8.1;
- Adds missing types where there were none at all, or was impossible to use because of unions;
- Adds Task object for type safety to be returned instead of undocumented array shapes;
- Fully enables `trailing_comma_in_multiline` cs fixer rule;
- Disables `php_unit_strict` cs fixer rule, because it breaks tests; 

**This introduces a BC break for someone who is extending meilisearch classes, but not for those who consumes the data, as Task object was added implementing `\ArrayAccess`, so all raw data can still be accessed.**

## PR checklist


tacman:
YAY!

If these are to be deprecated, are they going to be replaced with direct access to the object?

```php
    public function getData(): array
    {
        return $this->data;
    }

    // @todo: deprecate
    public function offsetExists(mixed $offset): bool
    {
        return \array_key_exists($offset, $this->data);
    }

    // @todo: deprecate
    public function offsetGet(mixed $offset): mixed
    {
        return $this->data[$offset] ?? null;
    }

    // @todo: deprecate
    public function offsetSet(mixed $offset, mixed $value): void
    {
        throw new \LogicException(\sprintf('Setting data on "%s::%s" is not supported.', get_debug_type($this), $offset));
    }

    // @todo: deprecate
    public function offsetUnset(mixed $offset): void
    {
        throw new \LogicException(\sprintf('Unsetting data on "%s::%s" is not supported.', get_debug_type($this), $offset));
    }
```

norkunas:
@tacman if this PR will get merged, then you'll get access to both ways: `$task->getError()` or `$task['error']`
but imho previous way should be deprecated, because it does not provide types

Strift:
Hi, thanks for your PR. 

Do we want to release this in a v2.x? 

Can you add a "migration guide" to your first comment to highlight the required implementation changes?

norkunas:
@Strift

> Do we want to release this in a v2.x?

Based on semver, it should be in v2.x, but if you want to be in line with [meilisearch/meilisearch](https://github.com/meilisearch/meilisearch) repository, I think it's reasonable to release this in v1.

> Can you add a "migration guide" to your first comment to highlight the required implementation changes?

Can you link to some previous comment where I could see how this looks like?

Strift:
Hi @norkunas, no need to follow a specific format. A simple before/after of the relevant parts of the code is enough. Here's an example for meilisearch JS: https://github.com/meilisearch/meilisearch-js/releases/tag/v0.48.0 

It's mainly so I have examples to refer to when writing the release changelog. Double-checking with the actual code changes is better, but the code samples are a good resource too for me, especially if I write the changelog a long time after reviewing the PR.

norkunas:
@Strift I've updated description, is it ok for you?
Also it should be decided if we deprecate the array access or not.

norkunas:
I also have a side idea for another improvement:
Add `wait(int $timeoutInMs, int $intervalInMs)` method to task.
So instead of:
```php
$task = $this->client->createIndex('new-index', ['primaryKey' => 'objectID']);
$task = $this->index->waitForTask($task->getTaskUid());
```

we could write:
```php
$task = $this->client->createIndex('new-index', ['primaryKey' => 'objectID'])->wait();
```

But not sure, as then we'd need for this method to work at least inject in in the Task object the `Meilisearch\Endpoints\Tasks` endpoint, or the http client, or the meilisearch client itself..

norkunas:
Hi, I'm for merging to v1 :) I think this is an acceptable BC break to provide type safety

norkunas:
I also think that it would be good to have typed task `details` based on https://www.meilisearch.com/docs/reference/api/tasks#details, because now it's always `array<mixed>`, so static analyzers will always complain.

src/Endpoints/Delegates/HandlesSystem.php
```diff
@@ -47,7 +48,7 @@ public function generateTenantToken(string $apiKeyUid, $searchRules, array $opti
         return $this->tenantToken->generateTenantToken($apiKeyUid, $searchRules, $options);
     }
 
-    public function swapIndexes(array $indexes): array
+    public function swapIndexes(array $indexes): Task
```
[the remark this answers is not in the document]
norkunas: @Copilot but it shouldn't, because `Indexes::swapIndexes` wraps it

tests/Settings/PaginationTest.php
```diff
@@ -0,0 +1,51 @@
...
+    public function testUpdatePagination(): void
+    {
+        $promise = $this->index->updatePagination(['maxTotalHits' => 100]);
+
+        $this->index->waitForTask($promise['taskUid']);
+
+        self::assertSame(['maxTotalHits' => 100], $this->index->getPagination());
+    }
+
+    public function testResetRankingRules(): void
```
Strift, line 40:
```suggestion
    public function testResetPagination(): void
```

Strift (changes requested):
This mostly looks good to me. I left a few comments.

I would love to release this with the changes related to Meilisearch v1.15.

Also, Codecov says the following lines are not covered by tests:
- src/Contracts/Task.php#L114
- src/Contracts/Task.php#L116

Strift:
> I also have a side idea for another improvement:
> Add wait(int $timeoutInMs, int $intervalInMs) method to task.

This could be a welcome addition, but I prefer having it in a separate PR to keep this one's scope minimal.

> I also think that it would be good to have typed task details based on https://www.meilisearch.com/docs/reference/api/tasks#details, because now it's always array<mixed>, so static analyzers will always complain.

Can we tackle this in another PR? We can create an issue to track it in the meantime.

norkunas:
Thanks @Strift for review, I'm currently preparing for `details` data to be typed.

src/Contracts/TaskType.php
```diff
@@ -0,0 +1,21 @@
...
+
+enum TaskType: string
+{
+    case IndexCreation = 'indexCreation';
+    case IndexUpdate = 'indexUpdate';
+    case IndexDeletion = 'indexDeletion';
+    case IndexSwap = 'indexSwap';
+    case DocumentAdditionOrUpdate = 'documentAdditionOrUpdate';
+    case DocumentDeletion = 'documentDeletion';
+    case DocumentEdition = 'documentEdition';
```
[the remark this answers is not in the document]
norkunas: Will remove

src/Endpoints/Delegates/HandlesDocuments.php
```diff
@@ -6,15 +6,14 @@
...
 use Meilisearch\Exceptions\ApiException;
-use Meilisearch\Exceptions\InvalidArgumentException;
 use Meilisearch\Exceptions\InvalidResponseBodyException;
 
 trait HandlesDocuments
 {
-    public function getDocument($documentId, ?array $fields = null)
+    public function getDocument(string|int $documentId, ?array $fields = null): array
     {
-        $this->assertValidDocumentId($documentId);
```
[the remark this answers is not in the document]
norkunas: Forgot to add `@param non-empty-string` :)

tests/Contracts/TaskTest.php
```diff
@@ -0,0 +1,112 @@
...
+        self::assertSame(123, $task->getCanceledBy());
+        self::assertSame(666, $task->getBatchUid());
+        self::assertSame(['some' => 'detail'], $task->getDetails());
+        self::assertSame([
+            'message' => 'Index `documents` not found.',
+            'code' => 'index_not_found',
+            'type' => 'invalid_request',
+            'link' => 'https://docs.meilisearch.com/errors#index_not_found',
+        ], $task->getError());
+
```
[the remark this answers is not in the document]
norkunas: I don't think it's worth to move. Either we'll remove the test later or just this part from current test, so it's the same

src/Contracts/TaskType.php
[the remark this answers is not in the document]
norkunas: Oh now I see that it was used in tests which broke after I removed it.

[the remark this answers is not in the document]
norkunas: @Strift as you can see meilisearch docs are outdated: https://github.com/meilisearch/meilisearch-php/actions/runs/15012425388/job/42183478341?pr=735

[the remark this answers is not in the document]
Strift: I will report it, thanks!

src/Contracts/TaskDetails/IndexUpdateDetails.php
```diff
@@ -0,0 +1,30 @@
...
+        public readonly ?string $primaryKey,
+    ) {
+    }
+
+    public static function fromArray(array $data): self
+    {
+        return new self(
+            $data['primaryKey'],
+        );
+    }
```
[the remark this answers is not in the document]
norkunas: Why? it's not optional field

src/Contracts/TaskDetails/IndexCreationDetails.php
```diff
@@ -0,0 +1,30 @@
...
+        public readonly ?string $primaryKey,
+    ) {
+    }
+
+    public static function fromArray(array $data): self
+    {
+        return new self(
+            $data['primaryKey'],
+        );
+    }
```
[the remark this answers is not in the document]
norkunas: it's always available

.github/workflows/pre-release-tests.yml
```diff
@@ -88,39 +84,3 @@ jobs:
...
-
-      - name: Switch to Guzzle 6
-        run: |
-          sed -i 's/"guzzlehttp\/guzzle": "^[0-9]\+\.[0-9]\+\.[0-9]\+\",/"php-http\/guzzle6-adapter": "^2.0.2",/' composer.json
-
-      - name: Install dependencies
-        uses: ramsey/composer-install@v3
-
-      - name: Run test suite - php-http/guzzle6-adapter
-        run: sh scripts/tests.sh
```
brunoocasali, line 126: OMG thats beautiful 👏👏

.php-cs-fixer.dist.php
```diff
@@ -15,9 +15,8 @@
         'void_return' => true,
         'native_function_invocation' => ['include' => ['@compiler_optimized'], 'scope' => 'namespaced'],
         'php_unit_test_case_static_method_calls' => ['call_type' => 'self'],
-        'php_unit_strict' => true,
-        // @todo: when we'll support only PHP 8.0 and upper, we can enable `parameters` for `trailing_comma_in_multiline` rule
-        'trailing_comma_in_multiline' => ['after_heredoc' => true, 'elements' => ['array_destructuring', 'arrays', 'match'/* , 'parameters' */]],
+        'php_unit_strict' => false,
```
brunoocasali, line 18: Why changing this?

src/Contracts/Task.php
```diff
@@ -0,0 +1,197 @@
...
+        private readonly ?string $indexUid,
+        private readonly TaskStatus $status,
+        private readonly TaskType $type,
+        private readonly \DateTimeImmutable $enqueuedAt,
+        private readonly ?\DateTimeImmutable $startedAt = null,
+        private readonly ?\DateTimeImmutable $finishedAt = null,
+        private readonly ?string $duration = null,
+        private readonly ?int $canceledBy = null,
+        private readonly ?int $batchUid = null,
+        private readonly ?TaskDetails $details = null,
```
brunoocasali, line 37: When we introduce a new task type in the engine does will this code break?

tests/Endpoints/BatchesTest.php
```diff
@@ -36,7 +36,7 @@ public function testGetAllBatchesWithIndexUidFilters(): void
     public function testGetAllBatchesWithTasksFilters(): void
     {
         $tasks = $this->client->getTasks(new TasksQuery())->getResults();
-        $response = $this->client->getBatches((new BatchesQuery())->setUids([$tasks[0]['uid']]));
```
brunoocasali, line 39: Although now this is an object since the task implements `ArrayAccess`, I can still do the `['uid']` right?

tests/Settings/SearchableAttributesTest.php
```diff
@@ -28,26 +28,19 @@ public function testUpdateSearchableAttributes(): void
...
-        self::assertSame($searchableAttributes, $updatedAttributes);
+        self::assertSame($searchableAttributes, $indexA->getSearchableAttributes());
     }
 
     public function testResetSearchableAttributes(): void
     {
         $index = $this->createEmptyIndex($this->safeIndexName('books-1'));
-        $promise = $index->resetSearchableAttributes();
-
-        $this->assertIsValidPromise($promise);
```
brunoocasali, line 46: Thanks for removing this helper method 🙏

brunoocasali:
@norkunas After further discussing the subject with @curquiza, we decided it would be more prudent to release it as v2 since it introduces a few breaking changes, especially the deprecation of PHP 7.

If that becomes a burden for our users later, we can introduce a table to showcase which SDK version supports which Meilisearch version. In any case, I think that's better, the engine and their SDKs are independent pieces of software that should be able to evolve independently and should remain adherent to semver rules :) 

So, since we are doing a v2, feel free to suggest other potential breaking changes so we can release it in the same package!

Thanks a lot for the PR by the way, loved it! 

CC: @Strift

norkunas:
> @norkunas After further discussing the subject with @curquiza, we decided it would be more prudent to release it as v2 since it introduces a few breaking changes, especially the deprecation of PHP 7.
> 
> If that becomes a burden for our users later, we can introduce a table to showcase which SDK version supports which Meilisearch version. In any case, I think that's better, the engine and their SDKs are independent pieces of software that should be able to evolve independently and should remain adherent to semver rules :) 
> 
> So, since we are doing a v2, feel free to suggest other potential breaking changes so we can release it in the same package!
> 
> Thanks a lot for the PR by the way, loved it! 
> 
> CC: @Strift 

Sounds good 😉

norkunas:
Then I think we can remove array access of task data here?

src/Contracts/TaskDetails/TaskDeletionDetails.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public static function fromArray(array $data): self
+    {
+        return new self(
+            $data['matchedTasks'],
+            $data['deletedTasks'],
+            $data['originalFilter'],
+        );
+    }
```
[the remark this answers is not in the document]
norkunas: These keys are always available

tests/Endpoints/BatchesTest.php
norkunas: Yup, but now totally removed

.php-cs-fixer.dist.php
norkunas: Because it converts some call from `self::assertEquals` to `self::assertSame` and then the tests break

src/Contracts/Task.php
norkunas: Yes, this lib must be prepared to accept new task types/statuses/task details on new meilisearch releases, same as with other features, that @Strift or others adds when meilisearch gets them :)

brunoocasali:
> Then I think we can remove array access of task data here?

If you think we have more benefits by removing it than by keeping, I agree in removing it.

src/Contracts/Task.php
brunoocasali: But when I mean "break," will the user really see a parsing error or something like this if the type is not supported?

brunoocasali: I'm unsure if this is enough when we decided internally to move Meilisearch (the engine) to v1. One of the requirements for the integrations would be to be able to keep updating the engine no matter which version you have in the SDK.

So far, we've been successful with that promise if I'm not mistaken 😅. Would it be possible to handle it in a way we can redirect the code to a non-typed task type? So if a user updates Meilisearch, it won't break for them, and to get the full capabilities from the SDK, it will just require them to later upgrade the SDK, WDYT?

norkunas: But any change in api could break this sdk, so..

norkunas: > One of the requirements for the integrations would be to be able to keep updating the engine no matter which version you have in the SDK.

if that's a requirement, then we can resort back to using strings, which is sad, but that's your decision

norkunas:
@Strift i should switch to v2 branch target here?

norkunas:
@Strift can you upmerge to v2, because it requires changes here that are already done on v1.15

Strift:
Hi @norkunas, yes! Thanks for updating the target branch to `v2`.

brunoocasali:
So, I think we need to make the tests pass and merge this branch :)

Can you do this @norkunas then ask for my review again, please?

norkunas:
@brunoocasali that's not up to me. v2 needs to be upmerged, because `main` has some pushed changes

Strift:
Apologies, I may have made a mistake.

I merged the changes from v2 into this branch, but maybe I should have rebased the PR instead.

norkunas:
I never do merges

Strift:
Can we revert these 4 commits so we can start again?

I don't want to force push over your branch

norkunas:
@Strift i've removed those commits. i can now rebase, but the PR target is still v2, i should change it to main, which is now v2, right?

tacman:
Let's us know when we can start testing V2.  I'm eager to use the new return types.

Strift:
@norkunas yes, you can target `main` directly

Strift:
@coderabbitai review

src/Contracts/TaskType.php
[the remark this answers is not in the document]
Strift: Fixed in https://github.com/meilisearch/documentation/issues/3255

src/functions.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Meilisearch;
+
+/**
+ * @internal
+ */
```
Strift, line 17:
```suggestion
/**
 * @internal
 * 
 * Creates a partially applied function by binding initial arguments to the given callable. 
 * Returns a Closure that, when invoked, calls the original callable with the bound arguments 
 * prepended to any new ones. 
 * 
 * Used internally to build reusable “waiter” functions (e.g., binding the HTTP client to 
 * task-waiting logic) and reduce repetitive argument passing.
 */
```

Strift, line 17:
```suggestion
/**
 * @internal
 * 
 * Creates a partially applied function by binding initial arguments to the given callable. 
 * 
 * Returns a Closure that, when invoked, calls the original callable with the bound arguments prepended
 * to any new ones. 
 * 
 * Used internally to build reusable “waiter” functions (e.g., binding the HTTP client to 
 * task-waiting logic) and reduce repetitive argument passing.
 */
```

tests/Endpoints/IndexTest.php
```diff
@@ -185,56 +186,47 @@ public function testGetTasks(): void
...
-        self::assertArrayHasKey('type', $response);
-        self::assertSame('documentAdditionOrUpdate', $response['type']);
-        self::assertArrayHasKey('duration', $response);
-        self::assertArrayHasKey('enqueuedAt', $response);
-        self::assertArrayHasKey('startedAt', $response);
-        self::assertArrayHasKey('finishedAt', $response);
+        $task = $this->index->addDocuments([['id' => 1, 'title' => 'Pride and Prejudice']])->wait(750, 20);
+
+        self::assertSame(TaskStatus::Succeeded, $task->getStatus());
+        self::assertSame($task->getTaskUid(), $task->getTaskUid());
```
Strift:
```suggestion
```

tests/Endpoints/IndexTest.php
```diff
@@ -185,56 +186,47 @@ public function testGetTasks(): void
...
-        self::assertArrayHasKey('type', $response);
-        self::assertSame('documentAdditionOrUpdate', $response['type']);
-        self::assertArrayHasKey('duration', $response);
-        self::assertArrayHasKey('enqueuedAt', $response);
-        self::assertArrayHasKey('startedAt', $response);
-        self::assertArrayHasKey('finishedAt', $response);
+        $task = $this->index->addDocuments([['id' => 1, 'title' => 'Pride and Prejudice']])->wait(1000);
+
+        self::assertSame(TaskStatus::Succeeded, $task->getStatus());
+        self::assertSame($task->getTaskUid(), $task->getTaskUid());
```
Strift:
```suggestion
```

tests/Endpoints/IndexTest.php
```diff
@@ -184,56 +186,47 @@ public function testGetTasks(): void
...
-        self::assertSame($response['uid'], $promise['taskUid']);
-        self::assertArrayHasKey('type', $response);
-        self::assertSame('documentAdditionOrUpdate', $response['type']);
-        self::assertArrayHasKey('duration', $response);
-        self::assertArrayHasKey('startedAt', $response);
-        self::assertArrayHasKey('finishedAt', $response);
+        $task = $this->index->addDocuments([['id' => 1, 'title' => 'Pride and Prejudice']])->wait();
+
+        self::assertSame(TaskStatus::Succeeded, $task->getStatus());
+        self::assertSame($task->getTaskUid(), $task->getTaskUid());
```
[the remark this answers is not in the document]
Strift: @norkunas I agree and would remove this assertion

```suggestion
```

tests/Endpoints/DocumentsTest.php
```diff
@@ -6,38 +6,38 @@
...
+use Tests\MockTask;
 use Tests\TestCase;
 
 final class DocumentsTest extends TestCase
 {
     public function testAddDocuments(): void
     {
         $index = $this->createEmptyIndex($this->safeIndexName('movies'));
-        $promise = $index->addDocuments(self::DOCUMENTS);
+        $task = $index->addDocuments(self::DOCUMENTS)->wait();
```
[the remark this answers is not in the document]
Strift:
```suggestion
        $index->addDocuments(self::DOCUMENTS)->wait();
```

tests/Endpoints/DocumentsTest.php
```diff
@@ -53,10 +53,7 @@ public function testAddDocumentWithSpecialChars(): void
         ];
 
         $index = $this->createEmptyIndex($this->safeIndexName('movies'));
-        $promise = $index->addDocuments($documents);
-
-        $this->assertIsValidPromise($promise);
-        $index->waitForTask($promise['taskUid']);
+        $task = $index->addDocuments($documents)->wait();
```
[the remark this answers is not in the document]
Strift:
```suggestion
        $index->addDocuments($documents)->wait();
```

tests/Endpoints/DocumentsTest.php
```diff
@@ -216,17 +202,16 @@ public function testGetMultipleDocumentsByIds(): void
     public function testReplaceDocuments(): void
     {
         $index = $this->createEmptyIndex($this->safeIndexName('movies'));
-        $response = $index->addDocuments(self::DOCUMENTS);
-        $index->waitForTask($response['taskUid']);
+
+        $task = $index->addDocuments(self::DOCUMENTS)->wait();
```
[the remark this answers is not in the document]
Strift:
```suggestion
        $index->addDocuments(self::DOCUMENTS)->wait();
```

tests/Endpoints/DocumentsTest.php
```diff
@@ -298,9 +280,10 @@ public function testUpdateDocumentsByFunction(): void
     {
         $http = new Client($this->host, getenv('MEILISEARCH_API_KEY'));
         $http->patch('/experimental-features', ['editDocumentsByFunction' => true]);
+
         $index = $this->createEmptyIndex($this->safeIndexName('movies'));
-        $documentPromise = $index->addDocuments(self::DOCUMENTS);
-        $index->waitForTask($documentPromise['taskUid']);
+
+        $task = $index->addDocuments(self::DOCUMENTS)->wait();
```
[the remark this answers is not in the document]
Strift:
```suggestion
        $index->addDocuments(self::DOCUMENTS)->wait();
```

tests/Endpoints/DocumentsTest.php
```diff
@@ -429,17 +411,16 @@ public function testAddDocumentsNdjsonInBatches(): void
     public function testAddWithUpdateDocuments(): void
     {
         $index = $this->createEmptyIndex($this->safeIndexName('movies'));
-        $response = $index->addDocuments(self::DOCUMENTS);
-        $index->waitForTask($response['taskUid']);
+
+        $task = $index->addDocuments(self::DOCUMENTS)->wait();
```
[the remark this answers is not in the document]
Strift:
```suggestion
        $index->addDocuments(self::DOCUMENTS)->wait();
```

norkunas:
I'll try to find some time to continue working on this, but currently my primary work consumes all my time

src/Contracts/SimilarDocumentsQuery.php
```diff
@@ -7,9 +7,9 @@
 class SimilarDocumentsQuery
 {
     /**
-     * @var int|string
+     * @var int|non-empty-string
      */
-    private $id;
+    private int|string $id;
```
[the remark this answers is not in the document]
norkunas: This is a contract, if someone does not follow, then it's on them. Static analysis should warn them

norkunas:
@Strift enums basically are already value objects, so I don't see a point in wrapping them.
I've added `Unknown` values to enums itself. Can you please recheck the PR? I've rebased to fix conflicts (oh boy, it took some time.. :D), i hope this will be okay for you

norkunas:
Btw `integration-tests (PHP 8.4)` and `integration-tests (PHP 8.0)` should be removed from repo settings as required jobs.

src/Contracts/Task.php
```diff
@@ -0,0 +1,188 @@
...
+                TaskType::DocumentDeletion => null !== $details ? DocumentDeletionDetails::fromArray($details) : null,
+                TaskType::DocumentEdition => null !== $details ? DocumentEditionDetails::fromArray($details) : null,
+                TaskType::SettingsUpdate => null !== $details ? SettingsUpdateDetails::fromArray($details) : null,
+                TaskType::DumpCreation => null !== $details ? DumpCreationDetails::fromArray($details) : null,
+                TaskType::TaskCancelation => null !== $details ? TaskCancelationDetails::fromArray($details) : null,
+                TaskType::TaskDeletion => null !== $details ? TaskDeletionDetails::fromArray($details) : null,
+                // It’s intentional that SnapshotCreation tasks don’t have a details object
+                // (no SnapshotCreationDetails exists and tests don’t exercise any details)
+                TaskType::SnapshotCreation => null,
+                TaskType::Unknown => UnknownTaskDetails::fromArray($details),
```
Strift: What if `$details` is null? The `UnknownTaskDetails::fromArray()` method expects an array

Strift (approved):
Great work! All things LGTM now.

I have a question regarding the parsing of task types/statuses. Should we add a PSR-3 warning? 

e.g. `"Unknown Meilisearch task status: <status>; mapped to TaskStatus::Unknown. Consider upgrading the SDK for typed support."`

src/Contracts/Task.php
norkunas: You are right, fixed.

norkunas:
> Great work! All things LGTM now.
> 
> I have a question regarding the parsing of task types/statuses. Should we add a PSR-3 warning?
> 
> e.g. `"Unknown Meilisearch task status: <status>; mapped to TaskStatus::Unknown. Consider upgrading the SDK for typed support."`

Well for now this library did not depend on the PSR-3, but sure we can adopt it if you want :) This can be for follow-up PR, with a new issue opened

Strift:
Thanks for this gigantic PR! 

I'm thrilled to merge it finally 🙌 congrats!

tacman:
Wahoo!  Congrats and thanks.

PR #757, opened 2025-05-29
Cast federation payload explicilty to an object

norkunas:
# Pull Request

## Related issue
Fixes #685
Replaces #686

## What does this PR do?
- It casts the federation payload explicitly to an object. This solves the error that appears when no MultiSearchFederation properties have been set.

First commit is cherry-picked to give @thijskuilman original credit

## PR checklist
Please check if your PR fulfills the following requirements:
- [x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted the multi-search request payload so that federation data is sent as an object rather than an array.

- **Tests**
  - Added a test to verify that federation data in multi-search requests is correctly serialized as an object.

PR #758, opened 2025-05-29
Run tests only with a single http client

norkunas:
# Pull Request

## Related issue
Fixes #<issue_number>

## What does this PR do?
- Removes testing with multiple http clients as it's not sdk responsibility to test if they behaves correctly. PSR18 clients are interoperable, so they must work same by contract.

## Why?
- Because it blocks writing simple unit tests where we don't need to call meilisearch service like https://github.com/meilisearch/meilisearch-php/pull/757 - currently I'd need to adapt test based on available http client

## PR checklist

## Summary by CodeRabbit

- **Chores**
  - Simplified test workflows by removing testing across multiple HTTP client implementations.
  - Removed jobs related to PHP 7.4 with Guzzle 6 from test workflows.
  - Updated development dependencies to replace Guzzle with Symfony HTTP client and added it as a suggested package.

### DIFF
--- a/.github/workflows/meilisearch-beta-tests.yml
+++ b/.github/workflows/meilisearch-beta-tests.yml
@@ -27,7 +27,7 @@ jobs:
   tests:
     runs-on: ubuntu-latest
     needs: ['meilisearch-version']
-    name: integration-tests-against-rc (PHP ${{ matrix.php-version }}) (${{ matrix.http-client }})
+    name: integration-tests-against-rc (PHP ${{ matrix.php-version }})
     services:
       meilisearch:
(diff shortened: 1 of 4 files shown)


Strift (commented):
I support this change, but would prefer to have feedback from @brunoocasali too before moving forward.

brunoocasali (dismissed):
From what I understand if we state that we are PSR compliant it means no matter which HTTP library the customer uses, it should work. And that's the beauty of it. I never understood why we had such complex CI in the first place LGTM!

brunoocasali (approved):

PR #777, opened 2025-08-18
fix: wrong type for `retrieveVectors`

bpolaszek:
Setting `retrieveVectors` to `true` using `DocumentsQuery` currently lead to the following error when using the `POST` verb to retrieve documents:

```
[Meilisearch\Exceptions\ApiException (400)]
  Invalid value type at `.retrieveVectors`: expected a boolean, but found a string: `"true"`
```

It should only be casted as string when using `GET /documents` (which is deprecated for some reason).

This PR fixes that.

## Summary by CodeRabbit

- New Features
  - Added ability to enable vector retrieval via a flag in document queries.

- Bug Fixes
  - Serialized query payloads now represent the vector retrieval flag as boolean or null, matching docs and improving interoperability.
  - Boolean query parameters are now encoded consistently in request URLs to ensure correct transmission.

- Tests
  - Added coverage validating the vector retrieval flag is set and serialized correctly.

### DIFF
--- a/src/Contracts/DocumentsQuery.php
+++ b/src/Contracts/DocumentsQuery.php
@@ -123,7 +123,7 @@ class DocumentsQuery
      *     limit?: non-negative-int,
      *     fields?: non-empty-list<string>|non-empty-string,
      *     filter?: list<non-empty-string|list<non-empty-string>>,
-     *     retrieveVectors?: 'true'|'false',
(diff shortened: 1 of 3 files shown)


bpolaszek:
Looks like something is broken on the existing tests ?

norkunas:
Tests are being fixed in #780

PR #780, opened 2025-08-20
Make index swaps tests more permissive

Strift:
Fix [Meilisearch v1.18 tests run](https://github.com/meilisearch/meilisearch/actions/runs/17061126416/job/48368042229)


norkunas:
Why not adding `rename => false` to expected structure instead?

Strift:
My initial goal was to make the tests more resilient to the API adding new fields, so I focused on this. 

You're right, though; I might address the `rename` field in this PR. Updated 🙏

PR #778, opened 2025-08-19
tests: increase PHP memory limit to 512M in Dockerfile

bpolaszek:
Hello there,

When running `docker-compose run --rm package bash -c "composer install && composer test && composer lint && composer phpstan"`  I got memory_limit errors during the phpstan step (Mac M4).

I suggest to raise the default limit from 128M to 512M to avoid these errors.

## Summary by CodeRabbit

* **New Features**
  * Composer is now preinstalled in the container, enabling dependency management directly within the image.

* **Chores**
  * PHP memory limit set to 512M to improve application stability under heavier workloads.

PR #790, opened 2025-09-18
Update composer command to run phpstan

Strift:
# Pull Request

## What does this PR do?
- Update memory limit for phpstan

## PR checklist

## Summary by CodeRabbit

- Chores
  - Updated development script configuration to run static analysis with a defined memory limit, improving reliability and consistency during developer workflows and CI runs. This helps prevent out-of-memory failures when analyzing larger codebases. No changes to user-facing functionality or behavior.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,7 +69,7 @@
         "lint:fix": [
             "./vendor/bin/php-cs-fixer fix --verbose --config=.php-cs-fixer.dist.php --using-cache=no  --diff"
         ],
-        "phpstan": "./vendor/bin/phpstan",
+        "phpstan": "./vendor/bin/phpstan --memory-limit=512M",
         "test": [
             "sh scripts/tests.sh"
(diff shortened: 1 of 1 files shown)


norkunas:
What's the point of this?

Strift:
I often run into memory issues when running `composer phpstan` locally. 

This PR addresses the same memory limit issue fixed in #778. That other PR only focused on the Docker compose environment, though.

I can circumvent it by running `./vendor/bin/phpstan --memory-limit=512M` instead, but I thought it would be better if the composer script contained the correct arguments.

What do you think?

norkunas:
But that depends on the environment. If i have much memory, why it should limit and run slower? :slightly_smiling_face: 

in #778 it was increased, so not the same

Strift:
Thanks for the clarification! Indeed, everyone should configure it manually if needed

PR #895, opened 2026-03-25
Add method to get task documents

### LINKED ISSUE #890: [Meilisearch v1.13.0] Add method to get tasks documents
## Context

Meilisearch 1.13 introduces a method to [get tasks documents](https://www.meilisearch.com/docs/reference/api/async-task-management/get-tasks-documents).

## Description

We want to add a method to this SDK to interact with the `GET /tasks/{task_id}/documents` API endpoint.

## Tasks

- [ ] Update API method to allow retrieving a tasks documents
- [ ] Add test cases for the new method
- [ ] Add an example under the `get_task_documents_1` YAML key in [.code-samples.meilisearch.yaml](../blob/main/.code-samples.meilisearch.yaml). It should match the example in [the documentation](https://github.com/meilisearch/documentation/blob/main/.code-samples.meilisearch.yaml).

mvanhorn:
## Summary

Adds `getTaskDocuments()` for the `GET /tasks/{task_id}/documents` endpoint introduced in Meilisearch v1.13.

## Changes

- `src/Endpoints/Tasks.php`: Added `getDocuments(int $taskUid)` that calls `GET /tasks/{id}/documents`
- `src/Endpoints/Delegates/HandlesTasks.php`: Added `getTaskDocuments(int $uid)` delegate method
- `.code-samples.meilisearch.yaml`: Added `get_task_documents_1` code sample

## Testing

- Follows the existing pattern from `getTask()` / `get()`
- Code sample matches the format from the [Meilisearch documentation](https://www.meilisearch.com/docs/reference/api/async-task-management/get-tasks-documents)

Fixes #890

This contribution was developed with AI assistance (Claude Code).

## Summary by CodeRabbit

* **New Features**
  * Retrieve documents associated with a specific task ID via a new endpoint method, returning the task’s document payload.

* **Documentation**
  * Added a code sample showing how to fetch documents for a given task.

* **Tests**
  * Added an integration test that enables the experimental route, seeds a completed task, and verifies fetching task documents returns a non-empty array.

### DIFF
--- a/tests/Endpoints/TasksTest.php
+++ b/tests/Endpoints/TasksTest.php
@@ -55,6 +55,7 @@ final class TasksTest extends TestCase
         $http->patch('/experimental-features', ['getTaskDocumentsRoute' => true]);
 
         $task = $this->index->updateDocuments(self::DOCUMENTS);
+        $task->wait();
 
         $stream = $this->client->getTaskDocuments($task->getTaskUid());
 
@@ -63,7 +64,6 @@ final class TasksTest extends TestCase
         self::assertNotEmpty($lines, 'Stream should contain at least one NDJSON line');
 
         $documents = array_map(fn (string $line) => json_decode($line, true, 512, \JSON_THROW_ON_ERROR), $lines);
-        self::assertNotEmpty($documents);
         self::assertArrayHasKey('id', $documents[0], 'Each document should have an id field');
     }
 
--- a/tests/Endpoints/TasksTest.php
+++ b/tests/Endpoints/TasksTest.php
@@ -48,6 +48,15 @@ final class TasksTest extends TestCase
         self::assertInstanceOf(DocumentAdditionOrUpdateDetails::class, $task->getDetails());
     }
 
+    public function testGetTaskDocumentsClient(): void
+    {
+        [$seedTask, $completedTask] = $this->seedIndex();
+
+        $documents = $this->client->getTaskDocuments($completedTask->getTaskUid());
+
+        self::assertIsArray($documents);
+    }
+
     public function testGetAllTasksClient(): void
     {
         $tasks = $this->client->getTasks();
--- a/tests/Endpoints/TasksTest.php
+++ b/tests/Endpoints/TasksTest.php
@@ -50,11 +50,10 @@ final class TasksTest extends TestCase
 
     public function testGetTaskDocumentsClient(): void
     {
-        [$seedTask, $completedTask] = $this->seedIndex();
+        [, $completedTask] = $this->seedIndex();
 
         $documents = $this->client->getTaskDocuments($completedTask->getTaskUid());
 
-        self::assertIsArray($documents);
         self::assertNotEmpty($documents);
     }
 
--- a/tests/Endpoints/TasksTest.php
+++ b/tests/Endpoints/TasksTest.php
@@ -13,6 +13,7 @@ use Meilisearch\Contracts\TaskStatus;
 use Meilisearch\Contracts\TaskType;
 use Meilisearch\Endpoints\Indexes;
 use Meilisearch\Exceptions\ApiException;
+use Meilisearch\Http\Client;
 use Tests\TestCase;
 
 final class TasksTest extends TestCase
@@ -50,6 +51,9 @@ final class TasksTest extends TestCase
 
     public function testGetTaskDocumentsClient(): void
     {
+        $http = new Client($this->host, getenv('MEILISEARCH_API_KEY'));
+        $http->patch('/experimental-features', ['getTaskDocumentsRoute' => true]);
+
         [, $completedTask] = $this->seedIndex();
 
         $documents = $this->client->getTaskDocuments($completedTask->getTaskUid());
--- a/tests/Endpoints/TasksTest.php
+++ b/tests/Endpoints/TasksTest.php
@@ -55,6 +55,7 @@ final class TasksTest extends TestCase
         $documents = $this->client->getTaskDocuments($completedTask->getTaskUid());
 
         self::assertIsArray($documents);
+        self::assertNotEmpty($documents);
     }
 
     public function testGetAllTasksClient(): void
--- a/tests/Endpoints/TasksTest.php
+++ b/tests/Endpoints/TasksTest.php
@@ -14,7 +14,6 @@ use Meilisearch\Contracts\TaskType;
 use Meilisearch\Endpoints\Indexes;
 use Meilisearch\Exceptions\ApiException;
 use Meilisearch\Http\Client;
-use Psr\Http\Message\StreamInterface;
 use Tests\TestCase;
 
 final class TasksTest extends TestCase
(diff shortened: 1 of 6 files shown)


norkunas:
Missing tests :)

mvanhorn:
Added test in 2173521: verifies `getTaskDocuments()` returns an array for a completed document addition task.

tests/Endpoints/TasksTest.php
```diff
@@ -48,6 +48,15 @@ public function testGetOneTaskClient(): void
         self::assertInstanceOf(DocumentAdditionOrUpdateDetails::class, $task->getDetails());
     }
 
+    public function testGetTaskDocumentsClient(): void
+    {
+        [$seedTask, $completedTask] = $this->seedIndex();
+
+        $documents = $this->client->getTaskDocuments($completedTask->getTaskUid());
+
+        self::assertIsArray($documents);
```
norkunas: Maybe asserting that array is not empty is better to prove that we really get something?

mvanhorn: Added `assertNotEmpty` in 4ac9ba2.

norkunas (approved):
Just fix CI

mvanhorn:
Fixed in 980d58b: removed the redundant `assertIsArray` (phpstan flagged it as `alreadyNarrowedType` since `getTaskDocuments()` already returns `array`) and dropped the unused `$seedTask` variable.

norkunas (approved):

Strift (changes requested):
Tests are failing: 

```
1) Tests\Endpoints\TasksTest::testGetTaskDocumentsClient
Meilisearch\Exceptions\ApiException: Getting the documents of an enqueued task requires enabling the `get task documents route` experimental feature. See https://github.com/orgs/meilisearch/discussions/808

/home/runner/work/meilisearch-php/meilisearch-php/src/Http/Client.php:248
/home/runner/work/meilisearch-php/meilisearch-php/src/Http/Client.php:171
/home/runner/work/meilisearch-php/meilisearch-php/src/Http/Client.php:76
/home/runner/work/meilisearch-php/meilisearch-php/src/Endpoints/Tasks.php:27
/home/runner/work/meilisearch-php/meilisearch-php/src/Endpoints/Delegates/HandlesTasks.php:25
/home/runner/work/meilisearch-php/meilisearch-php/tests/Endpoints/TasksTest.php:55
```

mvanhorn:
Enabled the experimental feature flag before the test call in 5f38a80 -- follows the same `$http->patch('/experimental-features', ...)` pattern used in DocumentsTest and other experimental endpoint tests.

tests/Endpoints/TasksTest.php
```diff
@@ -48,6 +49,18 @@ public function testGetOneTaskClient(): void
         self::assertInstanceOf(DocumentAdditionOrUpdateDetails::class, $task->getDetails());
     }
 
+    public function testGetTaskDocumentsClient(): void
+    {
+        $http = new Client($this->host, getenv('MEILISEARCH_API_KEY'));
+        $http->patch('/experimental-features', ['getTaskDocumentsRoute' => true]);
```
norkunas, line 55: @Strift as experimental features are the public/documented api, i think it should be available on `$this->client` like `$this->client->enableExperimentaFeature(...)`, so the tests wouldn't need to instantiate new http client, wdyt?

Strift: Yes, that would be better. But the `enableExperimentaFeature` method would be unstable, because the accepted set of features would vary as versions change.

We do it in the JS package, and I personally think that's fine.

norkunas: well we could just accept any string, so would not matter if list changes

mvanhorn:
Fixed the test in 96d3851. The issue was calling `getTaskDocuments()` on a completed task -- the API only returns documents for `enqueued` or `processing` tasks. Now calls it immediately after `updateDocuments()` before the task completes.

Re: the experimental feature enablement discussion -- happy to switch to `$this->client->enableExperimentalFeature(...)` if you settle on that API. Current approach follows the same raw HTTP pattern used in DocumentsTest.

norkunas:
@Strift this is somewhat unexpected that this endpoint returns ndjson :slightly_smiling_face:

mvanhorn:
Saw the NDJSON discussion -- should I update `getTaskDocuments()` to parse the NDJSON response (split lines + json_decode each), or would you rather handle the response format separately?

Strift:
Hey there,

I would like to align this feature with how it's being handled in our other SDKs (specifically the JavaScript one). The `/tasks/{task_id}/documents` endpoint returns **NDJSON** (Newline Delimited JSON), and I prefer to handle it via streaming rather than parsing the entire response into a standard JSON array. 

I prefer to align with what the Meilisearch team designed. Since it's an experimental method, we can still iterate if this doesn't work well.

To achieve this, I recommend the following changes:

1.  **Add `getStream` to the HTTP Contract**:
    In `src/Contracts/Http.php`, add the method definition:
    ```php
    /**
     * @throws ApiException
     */
    public function getStream(string $path, array $query = []): StreamInterface;
    ```

2.  **Implement `getStream` in the HTTP Client**:
    In `src/Http/Client.php`, implement the method using the existing `executeStream` helper:
    ```php
    public function getStream(string $path, array $query = []): StreamInterface
    {
        $request = $this->requestFactory->createRequest(
            'GET',
            $this->baseUrl.$path.$this->buildQueryString($query)
        );

        return $this->executeStream($request);
    }
    ```

3.  **Update the Tasks endpoint**:
    In `src/Endpoints/Tasks.php` and the `HandlesTasks` delegate, update the method to return a `StreamInterface` and use the new `getStream()` method. We should avoid converting the response to an array and instead return the raw NDJSON stream (similar to how `streamCompletion` works in `HandlesChatWorkspaceSettings`).

For reference, you can look at the equivalent implementation in the JavaScript SDK: [meilisearch-js PR #2156](https://github.com/meilisearch/meilisearch-js/pull/2156/changes).

mvanhorn:
Implemented in 902ee12:

- Added `getStream()` to `Http` contract and `Client` implementation
- `Tasks::getDocuments()` now returns `StreamInterface` via `getStream()`
- `HandlesTasks::getTaskDocuments()` return type updated to `StreamInterface`
- Test updated to assert `StreamInterface` and non-empty content

Followed the pattern from the existing `postStream`/`executeStream` pair. Happy to adjust if the JS SDK reference calls for anything different.

tests/Endpoints/TasksTest.php
```diff
@@ -48,6 +50,20 @@ public function testGetOneTaskClient(): void
...
+
+        $task = $this->index->updateDocuments(self::DOCUMENTS);
+
+        $stream = $this->client->getTaskDocuments($task->getTaskUid());
+
+        self::assertInstanceOf(StreamInterface::class, $stream);
+        $content = (string) $stream;
+        self::assertNotEmpty($content);
```
norkunas: well i'd like that the test would show how to access documents, because now it's unclear for me

mvanhorn: Updated in 81d91b5 - the test now parses the NDJSON stream line by line, decodes each document, and asserts the structure. Also fixes the phpstan `alreadyNarrowedType` error from the removed `assertInstanceOf`.

norkunas (approved):

mvanhorn:
Fixed in f289c8c: the test was calling `getTaskDocuments` before the task completed, causing the `JsonException`. Added `$task->wait()` before the fetch. Also removed the redundant `assertNotEmpty($documents)` that phpstan flagged.

mvanhorn:
Fixed in 38e2983. Two things were breaking this, found by reproducing against meilisearch-enterprise:latest:

1. `$task->wait()` let the task reach `succeeded`, but the route only returns documents for `enqueued`/`processing` tasks (hence "does not contain any documents"). The test now enqueues a large batch first so the inspected task stays queued while we read it.

2. The route sends `Content-Type: application/x-ndjson` but streams the objects concatenated with no newline separators (verified up to 2000 docs), so splitting on `\n` + `json_decode` with `JSON_THROW_ON_ERROR` couldn't parse it. The test now decodes the payload as a sequence of concatenated JSON values.

Verified: this test 12/12 across repeats, full TasksTest 10/10, php-cs-fixer + phpstan clean. @norkunas it now demonstrates real document access. The concatenated-no-newline stream is a bit awkward to consume - happy to move the parsing into the client if you'd prefer, but kept the raw stream to match the PR's current API.

Strift (approved):
LGTM 🙌

mvanhorn:
Thanks for reviewing and merging, @Strift - getTaskDocuments fills a gap the PHP SDK had versus the raw API.

PR #916, opened 2026-06-01
feat(stats): add query params for Meilisearch v1.44 showInternalDatabaseSizes and sizeFormat

### LINKED ISSUE #915: [Meilisearch v1.44.0] Add human-formatted sizes and detailed DB sizes in stats
## Context

[Meilisearch 1.44.0]https://github.com/meilisearch/meilisearch/releases/tag/v1.44.0) introduces human-formatted sizes and detailed DB sizes in stats.

## Objective

Update the corresponding SDK methods:
- [Get stats of index](https://www.meilisearch.com/docs/reference/api/indexes/get-stats-of-index)
- [Get stats of all indexes](https://www.meilisearch.com/docs/reference/api/stats)

Both related API endpoints now accept new query parameters:
- `showInternalDatabaseSizes`: boolean, optional, defaults to false. When present, the index stat objects in responses of the stat routes now contain an additional internalDatabaseSizes key, whose value is a dictionary of the internal database names and their current size in the index, like in the stats object of a batch.
- `sizeFormat`: human or raw: optional, defaults to raw. When present and set to human, then all database sizes in responses of the stat routes will be returned as a string containing an appropriate unit (MiB, GiB, etc). When missing or set to raw, then the current behavior of expressing the size in bytes as a number is retained.

**The keys in `internalDatabaseSizes` are subject to change and should not be exposed as a strongly typed object.**

## Tasks

- [ ] Add/update methods to handle interacting with the stats API endpoints
- [ ] Add/update test cases for the new/updated methods
- [ ] Update [.code-samples.meilisearch.yaml](../blob/main/.code-samples.meilisearch.yaml) to add examples under the following keys: `get_index_stats_1` and `get_indexes_stats_1`. It should match the examples in [the documentation](https://github.com/meilisearch/documentation/blob/main/.code-samples.meilisearch.yaml).

arieleli01212:
Closes #915.

Adds support for the two new query parameters introduced in [Meilisearch v1.44.0](https://github.com/meilisearch/meilisearch/releases/tag/v1.44.0) for the stats endpoints.

## Changes

**`src/Contracts/Stats.php`** — `databaseSize` and `usedDatabaseSize` widened from `int` to `int|string` to accommodate human-readable responses when `sizeFormat='human'`

**`src/Contracts/IndexStats.php`** — same widening for `rawDocumentDbSize` and `avgDocumentSize`; new optional `internalDatabaseSizes: ?array` field (present only when `showInternalDatabaseSizes=true`; keys subject to change per spec)

**`src/Endpoints/Stats.php`** — overrides `show()` to accept a `$query` array passed through to `http->get`

**`src/Endpoints/Delegates/HandlesSystem.php`** — `stats(array $query = [])` passes query to `Stats::show()`

**`src/Endpoints/Indexes.php`** — `stats(array $query = [])` passes query to `http->get`

**`.code-samples.meilisearch.yaml`** — updated `get_index_stats_1` and `get_indexes_stats_1`

All existing callers with no arguments continue to work unchanged.

## Summary

This pull request adds support for two new query parameters introduced in Meilisearch v1.44.0 to the stats endpoints: `showInternalDatabaseSizes` and `sizeFormat`.

### Key Changes

**Contracts Updated**
- src/Contracts/Stats.php: `databaseSize` and `usedDatabaseSize` widened from `int` to `int|string` to support both raw byte counts and human-readable strings.
- src/Contracts/IndexStats.php: `rawDocumentDbSize` and `avgDocumentSize` widened from `int` to `int|string`; added optional `internalDatabaseSizes: ?array` (map of internal DB names to sizes) and a `getInternalDatabaseSizes(): ?array` getter. `fromArray()` updated to accept the new field.

**Endpoints and Delegates Enhanced**
- src/Endpoints/Stats.php: added `show(array $query = []): ?array` to forward query parameters to the HTTP GET for `/stats`.
- src/Endpoints/Delegates/HandlesSystem.php: `stats()` signature changed to `stats(array $query = []): StatsContract` and forwards `$query` to `Stats::show()`.
- src/Endpoints/Indexes.php: `stats(array $query = []): array` now forwards the query array to the HTTP GET request for index stats.

**Documentation Samples**
- .code-samples.meilisearch.yaml: updated `get_index_stats_1` and `get_indexes_stats_1` to call `stats` with `showInternalDatabaseSizes: true` and `sizeFormat: 'human'`.

### Backward Compatibility

All additions are optional and default to prior behavior; callers that invoke `stats()` without arguments continue to work unchanged.

### Related Issues

Closes `#915`

### DIFF
--- a/src/Contracts/IndexStats.php
+++ b/src/Contracts/IndexStats.php
@@ -8,20 +8,22 @@ final class IndexStats
 {
     /**
      * @param non-negative-int                          $numberOfDocuments
-     * @param non-negative-int                          $rawDocumentDbSize
-     * @param non-negative-int                          $avgDocumentSize
+     * @param int|string                                $rawDocumentDbSize         Bytes or human-readable string
+     * @param int|string                                $avgDocumentSize           Bytes or human-readable string
      * @param non-negative-int                          $numberOfEmbeddings
      * @param non-negative-int                          $numberOfEmbeddedDocuments
      * @param array<non-empty-string, non-negative-int> $fieldDistribution
+     * @param array<non-empty-string, int|string>|null  $internalDatabaseSizes     Present when showInternalDatabaseSizes=true; keys subject to change
      */
     public function __construct(
         private readonly int $numberOfDocuments,
-        private readonly int $rawDocumentDbSize,
-        private readonly int $avgDocumentSize,
+        private readonly int|string $rawDocumentDbSize,
+        private readonly int|string $avgDocumentSize,
         private readonly bool $isIndexing,
         private readonly int $numberOfEmbeddings,
         private readonly int $numberOfEmbeddedDocuments,
         private readonly array $fieldDistribution,
+        private readonly ?array $internalDatabaseSizes = null,
     ) {
     }
 
@@ -34,17 +36,25 @@ final class IndexStats
     }
 
     /**
-     * @return non-negative-int
+     * Returns the raw document DB size.
+     * Value is an integer (bytes) when `sizeFormat` is `'raw'` (default),
+     * or a human-readable string such as `"2.3 MiB"` when `sizeFormat` is `'human'`.
+     *
+     * @return int|string
      */
-    public function getRawDocumentDbSize(): int
+    public function getRawDocumentDbSize(): int|string
     {
         return $this->rawDocumentDbSize;
     }
 
     /**
-     * @return non-negative-int
+     * Returns the average document size.
+     * Value is an integer (bytes) when `sizeFormat` is `'raw'` (default),
+     * or a human-readable string such as `"2.3 MiB"` when `sizeFormat` is `'human'`.
+     *
+     * @return int|string
      */
-    public function getAvgDocumentSize(): int
+    public function getAvgDocumentSize(): int|string
     {
         return $this->avgDocumentSize;
     }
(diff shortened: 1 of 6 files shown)


src/Endpoints/Indexes.php
```diff
@@ -243,9 +243,12 @@ public function facetSearch(FacetSearchQuery $params): FacetSearchResult
 
     // Stats
 
-    public function stats(): array
+    /**
+     * @param array{showInternalDatabaseSizes?: bool, sizeFormat?: 'raw'|'human'} $query
+     */
+    public function stats(array $query = []): array
```
norkunas, line 249: I'd use request object instead

src/Contracts/IndexStats.php
```diff
@@ -34,17 +36,25 @@ public function getNumberOfDocuments(): int
...
     /**
-     * @return non-negative-int
+     * Returns the raw document DB size.
+     * Value is an integer (bytes) when `sizeFormat` is `'raw'` (default),
+     * or a human-readable string such as `"2.3 MiB"` when `sizeFormat` is `'human'`.
+     *
+     * @return int|string
      */
-    public function getRawDocumentDbSize(): int
+    public function getRawDocumentDbSize(): int|string
```
norkunas, line 43: @Strift i'd instead tighten the return type to https://github.com/zenstruck/bytes, but not sure if team agrees on using external packages here (when i tried to implem stats command in meilisearch-bundle, i relied on this package, but the stats are not ported to v1 yet IIRC? so i didn't open a PR)
why? because then it's a consumers choice how to format to human readable format, what unit to use, etc.

Strift: This is mainly used for debugging. I don't think it's worth introducing a new dependency for now. Just passing through the data should be fine

norkunas: imho the api should have just returned a new field instead of introducing a boolean flag that just changes the response contract, but ok

PR #920, opened 2026-06-22
feat: support retrieveVectors in getDocument

### LINKED ISSUE #809: `getDocument` can't use `retrieveVectors`
**Description**
The documentation says you can get the vector data from a single document by passing `retrieveVectors` as well.

https://www.meilisearch.com/docs/reference/api/documents#get-one-document

**Expected behavior**
Pass `retrieveVectors: true` to `getDocument` and get the vector data

**Current behavior**
`getDocument` signature doesn't include `retrieveVectors`

```
public function getDocument($documentId, ?array $fields = null)
```
https://github.com/meilisearch/meilisearch-php/blob/main/src/Endpoints/Delegates/HandlesDocuments.php#L15C5-L15C68

tcferreira:
## Summary

The get-one-document endpoint accepts a `retrieveVectors` query parameter (to
include each document's vector data in the `_vectors` field), but
`Index::getDocument()` had no way to set it — only `getDocuments()` did, through
`DocumentsQuery`. This closes that gap. Closes #809.

## Changes

- `getDocument(string|int $documentId, ?array $fields = null, bool $retrieveVectors = false)`:
  when `true`, the request now sends `retrieveVectors=true`. New optional
  parameter, so it's non-breaking for existing callers.
- Integration test (`testGetDocumentWithVector`) mirroring the existing
  `testGetDocumentsWithVector`, asserting `_vectors` is absent by default and
  present when `retrieveVectors` is enabled.

## Overview
This PR adds support for the `retrieveVectors` query parameter to the `getDocument()` method in the Meilisearch PHP SDK, enabling retrieval of vector data for individual documents through the single-document endpoint. Previously, this functionality was only available through the `getDocuments()` method.

## Changes

**API Method Update**
- Modified `getDocument()` signature to accept an optional `$retrieveVectors` boolean parameter (defaults to `false` for backward compatibility)
- When `$retrieveVectors` is `true`, the method includes `retrieveVectors=true` in the query parameters sent to the `/documents/{id}` endpoint
- Refactored query parameter construction to conditionally build the request based on provided parameters

**Test Coverage**
- Added integration test `testGetDocumentWithVector()` that verifies:
  - The `_vectors` field is absent from the response by default
  - The `_vectors` field is included when `retrieveVectors` is enabled
  - Test uses a manual embedder with 3 dimensions and the `VECTOR_MOVIES` dataset

## Impact
This change brings feature parity between the single-document and multi-document retrieval methods, allowing developers to retrieve vector data via `getDocument()` just as they can with `getDocuments()`. Since the new parameter is optional with a default value of `false`, this is a non-breaking change for existing code.

### DIFF
--- a/src/Endpoints/Delegates/HandlesDocuments.php
+++ b/src/Endpoints/Delegates/HandlesDocuments.php
@@ -17,10 +17,17 @@ trait HandlesDocuments
 {
     /**
      * @param non-empty-string|int $documentId
+     * @param list<non-empty-string>|null $fields
      */
-    public function getDocument(string|int $documentId, ?array $fields = null): array
+    public function getDocument(string|int $documentId, ?array $fields = null, bool $retrieveVectors = false): array
     {
-        $query = isset($fields) ? ['fields' => implode(',', $fields)] : [];
+        $query = [];
+        if (isset($fields)) {
+            $query['fields'] = implode(',', $fields);
+        }
+        if ($retrieveVectors) {
+            $query['retrieveVectors'] = true;
+        }
 
         return $this->http->get(self::PATH.'/'.$this->uid.'/documents/'.$documentId, $query);
     }
--- a/tests/Endpoints/DocumentsTest.php
+++ b/tests/Endpoints/DocumentsTest.php
@@ -681,6 +681,22 @@ final class DocumentsTest extends TestCase
         self::assertCount(5, $response);
     }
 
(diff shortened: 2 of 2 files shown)


Strift (changes requested):
CI is not passing; please fix linter issues.

Probably no need to fix PHPStan issues, as they will be fixed by #925

